PR #13, opened 2015-08-29
add player font

mrzmyr:
Thats a little preparation for:

1. A build system (Babel & watch task)
2. Having icons like shuffle and repeat

I'm currently working on the repeat feature, if you'd like to have a PR for that as well, let me know.

screen shot 2015-08-29 at 15 55 05

Another vision could be, to replace awesome font with a own custom one in terms of reducing the effort of maintaining both.

PR #14, opened 2015-08-29
Add repeat

mrzmyr:
Hope you can see what happened here, because there are a lot of file changes in this PR.

Whats included in this PR:

- It depends on #13.
- It uses the [BEM methodology](https://en.bem.info/method/) for CSS naming
- Adds the repeat function (incl. repeat all, repeat one)
- Fixed up the header buttons using the new player font (e.g. repeat button)

I didn't used `<Button>` on purpose, because it was easier to overwrite the default styling of the normal `<button>` instead the Bootstrap defaults. Hope you are ok with that.

derp


KeitIG:
Hello !

I just created the shuffle feature, could you modify this PR to adapt to this:
capture

best !

mrzmyr:
May you want to merge your stuff first and I will merge it into this PR, so I can fix it up ?

KeitIG:
I already pushed that to origin master ;)

KeitIG:
One more thing, I just moved everything to gulp.

There is no need to `jsx --watch --extension blablabla` anymore. Just `gulp` in the app folder :)

KeitIG:
Also, could you fix that ? (in `gulpfile.js`)

```
gulp.task('font', function() {
    return gulp
        .src(['src/images/icons/*.svg'])
        .pipe(iconfont({
            fontName: 'player-font',
            appendUnicode: true,
            formats: ['ttf', 'eot', 'woff'],
            timestamp: Math.round(Date.now()/1000)
        }))
        .on('glyphs', function(glyphs, options) {
            var options = {
                glyphs: glyphs.map(function(glyph) {
                    return { name: glyph.name, codepoint: glyph.unicode[0].charCodeAt(0) };
                }),
                fontName: 'player-font',
                fontPath: '../fonts/',
                className: 'pf'
            };

            gulp.src('src/styles/templates/player-font.css')
                .pipe(consolidate('lodash', options))
                .pipe(gulp.dest('./src/dist/css/'));
        })
        .pipe(gulp.dest('src/fonts/'));
});
```

at the line `.on('glyphs', function(glyphs, options) {`, options is not defined, and then you define it in the callback function.

mrzmyr:
I merged and added the needed UI, but couldn't figure out why the click handler didn't work.
Can you have a quick look at it ?

KeitIG:
Oh I know why

KeitIG:
Great work !

I've read your code, I have a few things to say:

- [ ] for your click handler problem, try applying `-webkit-app-region: no-drag;` to the repeat button.
- [ ] why not keeping the <Button> from bootstrap-react ? (as it was for the shuffle button)

Finally, could you explain me why your class are like `now-playing__info-button now-playing__info-button--highlight` ? I find that a bit complex but I must have missed something

mrzmyr:
1. the `no-drag`fix worked, thanks for that
2. The bootstrap buttons kept making more overhead for overwriting css rules. I thought its way easier to use the native buttons.
3. Done :ship: 

The BEM methodology matches pretty good with the react ideas of classes and its a more consistent for further developers to join the project.

KeitIG:
Ok, working from origin add-repeat branch from now. I'm fixing stuff.

KeitIG:
Merging it now. On my way to fix the shuffle problem.

I'm not a big fan of the BEM methodology, for example:
capture
I find that does not work very well with SASS. I'll see about that later.

mrzmyr:
One goal of BEM is to flatten the cascading of CSS. In that way you win on readability and it becomes way more easy to structure.

Issue #57, opened 2015-12-17
Render tracks as tiles


KeitIG:
We're having problem wiith big libraries, TracksList render time is getting longer and longer, >3000 tracks make the render update noticeable with the eye.

We should implement what Atom has done with the editor view: render tracks in group (tiles), and only show the tiles needed when scrolling.

===

- [x] Tracks as tiles (absolute + translate3d transformation)
- [x] Show only needed tiles
- [x] Use `onScroll` on `.tracks-list-body` instead of `onWheel`
- [x] ~~Move `chunkArray` to Store to improve performance even more ?~~ (wontdo, not really needed, it takes 0.15ms)
- [x] Fix multi-select with maj.

===

helpers:
http://blog.atom.io/2015/06/24/rendering-improvements.html
http://blog.atom.io/2014/07/02/moving-atom-to-react.html

KeitIG:
#### First tests

Render time with 3200 tracks:
- before: 100ms (O(n) tome complexity) 
- after: 6ms (O(1) time complexity)

KeitIG:
All done.

PR #65, opened 2016-01-13
Tracks are now grouped in tiles, render time considerably improved (#57)

KeitIG:
#57 
for 0.2 release


KeitIG:
Let's merge this, we'll see the last bug later.

Issue #16, opened 2015-08-31
Split Sass in multiple files


KeitIG:
- Split styles in multiple files (see Webpack behavior with that)
- Use variables for colors
- ... bla

and all that stuff. Use SCSS just for nesting is ridiculous.

mrzmyr:
It would also be cool to make components out of the "one big file".

KeitIG:
Indeed !

mrzmyr:
Thats also a reason why I was preparing the components in the BEM methodology, that way its easier to move them to components.

KeitIG:
Still not the most urgent thing to do. One great thingwould be to compile SASS on the fly. A bit like Atom. This would make theme integration less dirty for example.

puranjayjain:
@KeitIG wouldn't it be easier to just switch to less and use less.js on the fly? Just a consideration (I'll be ready to help if a PR is needed :smile: )

KeitIG:
@puranjayjain Yes it would ! I was thinking about  that but I didn't really investigateurs how to do that. Feel free to submit a PR for that ❤

KeitIG:
Done in 97e6b8be2d4a7b651555b71eea1f48422d5e2067

PR #83, opened 2016-05-07
converted scss to less

puranjayjain:
Fixes towards a possible solution as discussed in #16 
* added lesshint for the sanity of the developers
* added less js for compiling on the fly

Reference Url:
[http://lesscss.org/#client-side-usage](http://lesscss.org/#client-side-usage)

Problems that i'm facing:
I don't know node-based build configs. I usually use gulp for that. 
So I need your help to add less.js injected in the app in the last and add the relevant less stylesheet in the head. 
There can be a small script which manages the theme like a thememanager which will siwtch theme based on user's preference. on runtime the default theme will be loaded anyway.


KeitIG:
Hello there !

First, thanks for this PR ! I was a bit tired when I answered you last friday, I'd like to keep Sass (and not switch to Less), and compile it on the fly with `node-sass` or sth, as Atom does with themes for example.

Sorry for your work, but I won't merge it. HOWEVER, there are some excellent things, that are imo necessary in a proach future:
- The styles on-the-fly compilation
- Something like the less linter you've implemented but for Sass
- Code consistency (color names ect...).

I'm opening a new issue for that, as the initial issue was not exactly this.

KeitIG:
#87

puranjayjain:
ok

PR #158, opened 2016-08-19
More natural volume + fix wrong volume on startup (fix #156)

### LINKED ISSUE #156: Add an option to use exponent function in volume control
Makes volume control match human preception

KeitIG:
poke @YurySolovyov: can you review it ?

Refs: #156

### DIFF
--- a/src/js/actions/PlayerActions.js
+++ b/src/js/actions/PlayerActions.js
@@ -62,7 +62,7 @@ export default {
 
             Player.setAudioVolume(volume);
 
-            app.config.set('audioVolume', volume);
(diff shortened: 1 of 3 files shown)


YurySolovyov:
I thought about using Math.exp here, but that would be a bit harder, cause it is not completely turns into 0. Can this hang a bit so I can test it when I'll back home?

Also, does this use any http://www.dr-lex.be/info-stuff/volumecontrols.html#about advices directly or you just feel that this ratio is good enough?

KeitIG:
See http://dr-lex.be/info-stuff/volumecontrols.html#ideal3

YurySolovyov:
LGTM, :ship:

PR #170, opened 2016-08-25
Fix wrong duration and wav errors

### LINKED ISSUE #103: A file may have no metadata
Something that could happen is a mp3 files without any metadata. Have to get that working too.

```
VM332:2 Uncaught Error: Could not find metadata header
```

===

### Todo

Basically, the goal is to make `utils.getMetadata` agnostic, whatever is the file sent to it, so we are sure to get nicely formatted metadata and duration

- [x] Get rid of mime in `utils.js` ?
- [ ] Create a `getDefaultMetadata` method in `utils.js` that is called when an error is thrown
- [ ] This method should return basically that:
``` javascript
// track is a file path
return {
    album        : 'Unknown',
    albumartist  : [],
    artist       : ['Unknown artist'],
    disk         : {
        no: 0,
        of: 0
    },
    duration     : duration, // must be computed somewhere
    genre        : [],
    loweredMetas : {},
    path         : track.path,
    playCount    : 0,
    title        : path.parse(track.path).base,
    track        : {
        no: 0,
        of: 0
    },
    type         : 'track',
    year         : ''
};
```
- [x] Call `getDefaultMetadata` directly for wav files
- [x] If musicmetadata does not return any duration, we should get it with something like (need perfs test). The problem is it's really slow.
``` javascript
const audio = new Audio();
audio.src = track.path;
audio.addEventListener('load', () => {
    // audio.duration
});
```
- [x] Maybe get rid of `wav-file-info` in that case

### LINKED ISSUE #145: Cannot Initialize Library
System: Win 10 64bit
Version: 0.6.0 (Downloaded from Webpage, extracted to homefolder)

Error in Dev-Console:
An error occured while reading G:/Media - Music/787f0f70 (2)/a1780a61-52ca3de8.mp3 id3 tags: Error: expected frame header but was not found

I added my Folder, klicked on "Refresh Library" and the progress-bar ran through (as the folder contains only around 1.7k files) and stopped at around 99%

as I have never developed on windows (using this system for gaming only) I cannot get inside of the server console.

KeitIG:
Fix #103 
Fix #145

### DIFF
--- a/src/js/utils/utils.js
+++ b/src/js/utils/utils.js
@@ -307,9 +307,10 @@ const utils = {
 
         audio.addEventListener('error', (e) => {
 
-            callback(`Error getting audio duration: (${e.target.error.code}) ${path}`, 0);
+            callback(new Error(`Error getting audio duration: (${e.target.error.code}) ${path}`), 0);
         });
 
+        audio.preload = 'metadata';
         audio.src = path;
     }
 };
--- a/src/js/utils/utils.js
+++ b/src/js/utils/utils.js
@@ -8,7 +8,6 @@ import path from 'path';
 import fs   from 'fs';
 
 import mmd     from 'musicmetadata';
-import wavInfo from 'wav-file-info';
 
 
 const utils = {
@@ -203,7 +202,7 @@ const utils = {
 
         if(path.extname(track) === '.wav') { // If WAV
 
-            wavInfo.infoByFilename(track, (err, info) => {
+            utils.getAudioDuration(track, (err, audioDuration) => {
 
                 if (err) console.warn(err);
 
@@ -216,7 +215,7 @@ const utils = {
                         no: 0,
                         of: 0
                     },
-                    duration     : info.duration,
+                    duration     : audioDuration,
                     genre        : [],
                     loweredMetas : {},
                     path         : track,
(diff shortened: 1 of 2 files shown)


src/js/utils/utils.js
```diff
@@ -274,9 +273,44 @@ const utils = {
...
+     *
+     * @param path (string)
+     * @return float
+     */
+    getAudioDuration(path, callback = () => {}) {
+
+        const audio = new Audio;
+
+        audio.addEventListener('loadedmetadata', () => {
+
```
YurySolovyov: I wonder if there is a way to stop loading rest of the file when we got metadata, say, would setting `src` to null do the job?

YurySolovyov:
Shouldn't we make all new APIs promise-returning?

src/js/utils/utils.js
```diff
@@ -274,9 +273,44 @@ const utils = {
...
+            callback(null, audio.duration);
+        });
+
+        audio.addEventListener('error', (e) => {
+
+            callback(`Error getting audio duration: (${e.target.error.code}) ${path}`, 0);
```
YurySolovyov: this should be new Error(...)

src/js/utils/utils.js
YurySolovyov: > The <audio> element includes a preload attribute. This can be set to "none" or "metadata" which should prevent the audio preloading.

See http://stackoverflow.com/a/36792488/1600598

KeitIG:
About promises: we're not forced to put promises everywhere, it's fine for now. If there's a callback hell here one day, we'll implement promises.

YurySolovyov:
I don't think it is about solving callback hell, it is about consistency

KeitIG:
Can you checkout the branch and make the needed changes ?

YurySolovyov:
Sure.

YurySolovyov:
You know, I tried making it Promise-based, but it does not look as nice as I thought it would.
How about landing this as is and creating an issue to rework all library-related functions to be Promise-based?

Branch is [`here`](https://github.com/YurySolovyov/museeks/tree/return-duration-promise) if you still want to look at it.

KeitIG:
Let it go, let it gooooo

Issue #176, opened 2016-08-26
Better media keys support + fix not working ones


YurySolovyov:
Electron can listen for [`globalShortcut`](http://electron.atom.io/docs/api/global-shortcut/)s via following [`Accelerator`](http://electron.atom.io/docs/api/accelerator/)s:
- MediaNextTrack, MediaPreviousTrack, MediaStop and MediaPlayPause
- VolumeUp, VolumeDown and VolumeMute

Let's use that.

KeitIG:
Hem.

https://github.com/KeitIG/museeks/blob/master/src/js/actions/AppActions.js#L121

YurySolovyov:
@KeitIG fair, volume still remains :)

KeitIG:
@Suriteka said the play/pause mac keyboard button was working, but not the next/prev, can you confirm that Jonathan ?

Suriteka:
@KeitIG Nop the Play/Pause don't work, just the Next and Previous works perfectly.

KeitIG:
ah ok, thanks I'll investigate

KeitIG:
@Suriteka @YurySolovyov 

I remember why I did not implement volume global shortcut: managing volume from the app does not make any sense, as if you toggle the mute key, it'll mute all your laptop, so no need to mute Museeks too. Same for `volume-up` and `volume-down`

YurySolovyov:
Ah, makes sense, well, at least you found the bug :)

KeitIG:
Done.

PR #177, opened 2016-08-26
Fix global shortcuts (#176)

KeitIG:


YurySolovyov:
Did it raised any exceptions without a fix?

KeitIG:
Nah. Exceptions is something we'll have to implement, I'm working on a pattern to deal with them across the app.

Issue #153, opened 2016-08-16
Make an option to run Museeks with native frame


KeitIG:
This would help to get minimize/maximize etc... (#41), the time we unblock the situation. In this case, remove the "close button"

- [x] Option "use native frame" in settings
- [x] Hide controls if this is checked
- [x] auto-restart the app ?

My only concern is with the dark theme on some Linux environments (libchromium being build with GTK2):

![capture du 2016-08-16 17 30 11](https://cloud.githubusercontent.com/assets/1311607/17705031/3dc86422-63d7-11e6-985e-afd2293d2c57.png)
![capture du 2016-08-16 17 30 25](https://cloud.githubusercontent.com/assets/1311607/17705030/3dc0ed0a-63d7-11e6-8de1-570359f9bb02.png)

YurySolovyov:
Here is the idea: make header render differenly depending on OS.

Render Windows-like buttons on the right for Windows:
![image](https://cloud.githubusercontent.com/assets/2168518/17715140/877e9760-640b-11e6-8338-3d34d8ade2b5.png)

Render OS X-like buttons on macOS:
![image](https://cloud.githubusercontent.com/assets/2168518/17715151/97b94d64-640b-11e6-8378-2b7f29f9912e.png)

Render some generic buttons on Linux: Same as today + minimize and maximize buttons

PR #179, opened 2016-08-26
153 native frame option

KeitIG:
#153

### DIFF
--- a/src/js/components/App.react.js
+++ b/src/js/components/App.react.js
@@ -47,6 +47,7 @@ class Museeks extends Component {
                     shuffle={ store.shuffle }
                     queue={ store.queue }
                     queueCursor={ store.queueCursor }
+                    config={ app.config.getAll() }
                 />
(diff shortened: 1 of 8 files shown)


src/js/components/App.react.js
```diff
@@ -47,6 +47,7 @@ class Museeks extends Component {
                     shuffle={ store.shuffle }
                     queue={ store.queue }
                     queueCursor={ store.queueCursor }
+                    config={ app.config.getAll() }
```
YurySolovyov: I think you can just pass `useNativeFrame` directly like:
```js
useNativeFrame: app.config.getAll().useNativeFrame
```

KeitIG:
Yes, but in case we have to work with more config options, I pref to it that way for now.

YurySolovyov:
I'd still prefer doing it that way, IMO passing config to component leaks unnecessary implementation details.

PR #188, opened 2016-08-29
173 large databases

KeitIG:
This is a drop of NeDB in favor or LinvoDB3.

It basically work like NeDB.

- Large database (hundred of Mo) is now working fine
- It is slightly more performant than NeDB when inserting documents
- It is extremly slower than NeDB to load the whole database if you have a lot of covers in your metatags (so we may find a solution for covers first)


src/js/components/Settings/LibraryFolders.react.js
```diff
@@ -11,7 +11,7 @@ import AppActions from '../../actions/AppActions';
 |--------------------------------------------------------------------------
 */
 
-export default class LibraryFolders extends PureComponent {
+export default class LibraryFolders extends Component {
```
YurySolovyov: I saw you mentioned this leads to error, any details?

YurySolovyov: so maybe do `Object.assign({}, app.config.getAll())` and try again?

KeitIG:
About this. I'm not sure how to proceed. I'd merge it as it's working fine with medium-large datasets (where NeDB failed), but it is really slow with covers, but I can open an issue for that as it's not directly related.

YurySolovyov:
Do you mean that getting cover is slow?
If It does not block the UI, merge it and make a ticket about covers.

KeitIG:
No, initial database load is slow (when you start the app)

YurySolovyov:
Then it is ok by me

PR #189, opened 2016-08-29
Minor UX improvements (fix #187)

### LINKED ISSUE #187: Trivial UX and user-feedback improvements
Before releasing 0.7.0, Museeks needs some refining:

- [x] Add a notification when tracks are succesfully added to a playlist
- [x] Add a link to Library from EmptyPlaylist
- ...

KeitIG:


YurySolovyov:
For now, the only user-facing change is that it adds new case for notifications, if that's it, you may want to change the title.

src/js/components/Playlists/EmptyPlaylist.react.js
```diff
@@ -1,25 +0,0 @@
...
-
-    render() {
-        return (
-            <FullViewMessage>
-                <p>Empty playlist</p>
-                <p className='sub-message'>You can add tracks from the library view</p>
-            </FullViewMessage>
-        );
-    }
-}
```
YurySolovyov: Can we do same thing we did for queues: make container own either empty playlist or non-empty one? Naming for this would be hard though...

KeitIG:
About componentize too much, look at: 

``` jsx
    getLibraryComponent() {
        if(this.props.library === null) {
            return <FullViewMessage><p>Loading library...</p></FullViewMessage>;
        }
        if (this.props.library.length === 0) {
            return <EmptyLibrary />;
        }
        if (this.props.tracks.length === 0) {
            return <FullViewMessage><p>Your search returned no results</p></FullViewMessage>;
        }
        return (
            <TracksList
                type='library'
                tracks={ this.props.tracks }
                trackPlayingId={ this.props.trackPlayingId }
                playlists={ this.props.playlists }
            />
        );
    }

    render() {
        return (
            <div className='view view-library' >
                { this.getLibraryComponent() }
            </div>
        );
    }
```

don't you feel it's just too much abstraction ? We're calling `FullViewMessage` multiple times, why make a separate component that renders just a new `FullViewMessage` ?

Imho, it really should be:

``` jsx
    getLibraryContent() {
        if(this.props.library === null) {
            return (
                <FullViewMessage>
                    <p>Loading library...</p>
                </FullViewMessage>
            );
        }
        if (this.props.library.length === 0) {
            return (
                <FullViewMessage>
                    <p>Too bad, there is no music in your library =(</p>
                    <p className='sub-message'>
                        <span>nothing found yet, but that's fine, you can always </span>
                        <Link to='/settings/library'>add your music here</Link>
                    </p>
                </FullViewMessage>
            );
        }
        if (this.props.tracks.length === 0) {
            return (
                <FullViewMessage>
                    <p>Your search returned no results</p>
                </FullViewMessage>
            );
        }
        return (
            <TracksList
                type='library'
                tracks={ this.props.tracks }
                trackPlayingId={ this.props.trackPlayingId }
                playlists={ this.props.playlists }
            />
        );
    }

    render() {
        return (
            <div className='view view-library' >
                { this.getLibraryContent() }
            </div>
        );
    }
```

And I'd put everying in the render method. But that's another thing, and I'm ok with calling class methods to get component content.

Abstracting things is a good idea when components gets too long or too complex, but not everwhere/everytime

YurySolovyov:
I agree that no abstraction is better than bad abstraction, but that does not mean we shouldn't *seek* for   a good abstractions.
We are not in a hurry, right? We don't *have to* ship anything, and can just sit back and think, right?

KeitIG:
Tbh, I don't really worry about how the code looks, because the codebase looks good anyway. I'm just a bit more focused on features for now.

KeitIG:
I think we broke appveyor :stuck_out_tongue_winking_eye:

YurySolovyov:
> Tbh, I don't really worry about how the code looks

Well, if you only start worrying when it looks bad, then it is a bit late.

KeitIG:
Well it's often late in that case ;)

I propose we pass once on the code before releasing, but now, let's focus on the features.

KeitIG:
About notifications ect, I think we can create a user-story/issue, even if ftm it's ok. I'd need some use-case first.

YurySolovyov:
>Well it's often late in that case ;)

Just wanted to make the point, nothing more.

> I propose we pass once on the code before releasing, but now, let's focus on the features.

and

> About notifications ect, I think we can create a user-story/issue, even if ftm it's ok. I'd need some use-case first.

Sounds good to me.

PR #195, opened 2016-09-04
Decent covers implementation

### LINKED ISSUE #166: Look for covers in track's directory
Tracking issue for making museeks look into track's directory for album cover art.

Questions to resolve
- [ ] what if track added from directory with >1 image there?
- [ ] what filenames to lookup ? all or just `/(cover|album|whatever).(jpg|png|bmp)/i` ?

Refs: #86

### LINKED ISSUE #191: Decent cover support
## Why

- Light cover support (#86) was ok, but definetly not scalable.
- #173 helped us to deal with large database, but the library loading is extremly long.

This issue is here to make things decent.

## What

- [x] Redefine tracks's cover schema
- [x] Do not store covers in base64 for each track (new collection, create images or look for them on the fly)
- [x] Look for covers in track directory #166 

## How

In my mind I have something like:

``` js
{ // track without any cover
    cover: null
},
{ // another track with a cover in its id3 tags
    type: 'id3',
    data: 'aCoverIdOrAbase64StringOrSthToBeDefined'
},
{ // another track with a cover.jpg in his directory
    type: 'file'
    data: 'filepath'
}
```

@YurySolovyov emmited the idea of an internal server serving them.

KeitIG:
- Fix #191: Decent cover support
- Fix #166: Look for covers in current track directory

### DIFF
--- a/src/js/utils/utils.js
+++ b/src/js/utils/utils.js
@@ -320,7 +320,7 @@ const utils = {
 
         const stream = fs.createReadStream(trackPath);
 
-        mmd(stream, { duration: true }, (err, data) => {
+        mmd(stream, (err, data) => {
 
             if(err) console.warn(err);
             else {
@@ -339,7 +339,7 @@ const utils = {
 
                 globby(pattern, { nodir: true, follow: false }).then((matches) => {
 
-                    const filteredMatches = matches.filter((elem) => {
+                    const cover = matches.find((elem) => {
 
                         const parsedPath = path.parse(elem);
 
@@ -347,12 +347,7 @@ const utils = {
                             && ['.png', '.jpg', '.bmp', '.gif'].includes(parsedPath.ext.toLowerCase()) ;
                     });
 
-                    if(filteredMatches[0]) {
-                        callback(filteredMatches[0]);
-                        return;
-                    }
-
-                    callback(null);
+                    callback(cover);
                     return;
                 });
             }
--- a/src/js/utils/utils.js
+++ b/src/js/utils/utils.js
@@ -6,8 +6,8 @@
 
 import path from 'path';
 import fs   from 'fs';
-
-import mmd     from 'musicmetadata';
+import mmd  from 'musicmetadata';
+import globby from 'globby';
 
 
 const utils = {
@@ -33,7 +33,6 @@ const utils = {
             result += `${minutes}:${seconds}`;
 
             return result;
-
         }
 
         return '00:00';
@@ -212,7 +211,6 @@ const utils = {
                     album        : 'Unknown',
                     albumartist  : [],
                     artist       : ['Unknown artist'],
-                    cover        : null,
                     disk         : {
                         no: 0,
                         of: 0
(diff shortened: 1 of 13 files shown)


src/js/utils/utils.js
```diff
@@ -312,6 +309,54 @@ const utils = {
...
+        if(!trackPath) {
+            callback(null);
+            return;
+        }
+
+        const stream = fs.createReadStream(trackPath);
+
+        mmd(stream, { duration: true }, (err, data) => {
+
+            if(err) console.warn(err);
```
YurySolovyov: maybe do a `return console.warn(err);` so we can drop 1 indent level?

src/js/utils/utils.js
```diff
@@ -312,6 +309,54 @@ const utils = {
...
+    fetchCover(trackPath, callback) {
+
+        if(!trackPath) {
+            callback(null);
+            return;
+        }
+
+        const stream = fs.createReadStream(trackPath);
+
+        mmd(stream, { duration: true }, (err, data) => {
```
YurySolovyov: why do we need duration here?

src/js/utils/utils.js
```diff
@@ -312,6 +309,54 @@ const utils = {
...
+                }
+
+                // scan folder for any cover image
+                const folder = path.dirname(trackPath);
+
+                const pattern = path.join(folder, '*');
+
+                globby(pattern, { nodir: true, follow: false }).then((matches) => {
+
+                    const filteredMatches = matches.filter((elem) => {
```
YurySolovyov: we should use `.find` here

src/js/utils/utils.js
KeitIG: My bad

src/js/utils/utils.js
KeitIG: I don't think It's a good idea: 
- the `return` statement should only return a string
- if `musicmetadata` fails, it should still scan the folder for a cover right ?

src/js/utils/utils.js
KeitIG: Indeed

src/js/utils/utils.js
YurySolovyov: > the return statement should only return a string

But it's a callback API, you pass the result to callback, so why care what to return?

> if musicmetadata fails, it should still scan the folder for a cover right ?

Ah, right.
This nesting still bothers me though. Promises?

src/js/utils/utils.js
```diff
@@ -312,6 +309,49 @@ const utils = {
...
+
+                    const cover = matches.find((elem) => {
+
+                        const parsedPath = path.parse(elem);
+
+                        return ['album', 'albumart', 'folder', 'cover'].includes(parsedPath.name.toLowerCase())
+                            && ['.png', '.jpg', '.bmp', '.gif'].includes(parsedPath.ext.toLowerCase()) ;
+                    });
+
+                    callback(cover);
```
YurySolovyov: what if we found nothing? Is code on the other side is ready to handle undefined?

KeitIG: It returns undefined, then https://github.com/KeitIG/museeks/pull/195/files#diff-6faf57b394e7c8c1a51f2c5c9f4cd5f5R421 handles it

YurySolovyov:
LGTM, Wanna do utils exports refactoring in a separate PR?

KeitIG:
Yes, I think it's preferable.

PR #196, opened 2016-09-04
Search albums and artists with a click.

commondream:
This is pretty rough, but puts together a quick concept of having albums and artists be links so clicking on them will run a search for them.

To do:

- [ ] Fill in search field after click. The input component being used right now makes that difficult.
- [ ] Add click to search to genre.

I'd love to hear feedback.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,9 +17,9 @@
     "ci:eslint": "./node_modules/.bin/eslint . --config .eslintrc",
     "build:compile": "./node_modules/.bin/webpack --config webpack.config.js",
     "build:eslint": "./node_modules/.bin/eslint . --config .eslintrc",
-    "compile": "webpack --config webpack.config.js --progress --colors",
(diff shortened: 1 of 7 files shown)


YurySolovyov:
> I'd love to hear feedback.

Well, first, thanks for the interest!
The thing is that we usually create issues to discuss adding features/changing behavior, that kind of stuff.

From the readme:

> Please respect a few rules:
- **Before making complex stuff, don't hesitate to open an issue first to discuss about it**
- Make the code readable and comment if needed
- Make sure npm run lint passes

KeitIG:
If I agree on the fact that it should be great to filter by artist or album, I disagree on the way it's done. Let discuss it here: #198

commondream:
I figured complex was an important word there. This only took a few minutes to spike, so throwing it away doesn't bother me at all.

KeitIG:
@commondream I'd like to get your feedbacks about #198 though :)

PR #201, opened 2016-09-11
Implement playCount database update

### LINKED ISSUE #160: Implement track playCount incrementation
There is a field for `{ type : 'tracks' }` called `playCount`. It would need to be incremented when needed. 

Will be used for: #62 
- [x] should it be incremented on `ended` event or when starting a track ?
- [ ] do it

cfollet:
Closes #160.

### DIFF
--- a/src/js/actions/AppActions.js
+++ b/src/js/actions/AppActions.js
@@ -34,6 +34,11 @@ const AppActions = {
         // Audio Events
         Player.getAudio().addEventListener('ended', AppActions.player.next);
         Player.getAudio().addEventListener('error', AppActions.player.audioError);
+        Player.getAudio().addEventListener('timeupdate', () => {
+            if (Player.isThresholdReached()) {
+                LibraryActions.incrementPlayCount(Player.getAudio().src);
+            }
+        });
         Player.getAudio().addEventListener('play', () => {
             ipcRenderer.send('playerAction', 'play');
         });
--- a/src/js/actions/LibraryActions.js
+++ b/src/js/actions/LibraryActions.js
@@ -179,5 +179,18 @@ export default {
                 cover
             });
         });
+    },
+
+    /**
+     * Update the play count attribute.
+     *
+     * @param source
+     */
+    incrementPlayCount(source) {
+        app.models.Track.update({ src: source }, { $inc: { playcount : 1 } }, (err) => {
+            if(err) {
+                console.warn(err);
+            }
+        });
     }
 };
--- a/src/js/lib/player.js
+++ b/src/js/lib/player.js
@@ -11,6 +11,9 @@ class Player {
         this.audio.playbackRate = options.playbackRate || 1;
         this.audio.volume = options.volume || 1;
         this.audio.muted = options.muted || false;
+
+        this.threshold = .75;
+        this.durationThresholdReached = false;
     }
 
     play() {
@@ -47,11 +50,20 @@ class Player {
     }
 
     setAudioSrc(src) {
+        // When we change song, need to update the thresholdReached indicator.
+        this.durationThresholdReached = false;
         this.audio.src = src;
     }
 
     setAudioCurrentTime(currentTime) {
         this.audio.currentTime = currentTime;
+    }
+
+    isThresholdReached() {
+        if(! this.durationThresholdReached && this.audio.currentTime >= this.audio.duration * this.threshold) {
+            this.durationThresholdReached = true;
+            return this.durationThresholdReached;
+        }
     }
 }
 


src/js/components/Header/PlayingBar.react.js
```diff
@@ -158,6 +159,14 @@ export default class PlayingBar extends Component {
 
     tick() {
         this.setState({ elapsed: Player.getAudio().currentTime });
+
+        if (!Player.isThresholdReached() && Player.getAudio().currentTime >= Player.getAudio().duration * Player.getThreshold()) {
```
YurySolovyov: I don't quite get the purpose of the `thresholdReached` variable and all associated infra.

cfollet: If we don't do that it will increment each tick() call when `Player.getAudio().currentTime >= Player.getAudio().duration * Player.getThreshold()` so basically every 100 ms after the threshold.

Plus, this also prevent [the incrementaion if the user scroll back](https://github.com/KeitIG/museeks/issues/160#issuecomment-245395483)

KeitIG:
Thanks for the PR, I'll review it a soon as I have some free time.

I'll try to investigate why AppVeyor fails too.

src/js/components/Header/PlayingBar.react.js
YurySolovyov: Maybe you can move `      Player.getAudio().currentTime >= Player.getAudio().duration * Player.getThreshold()` into local variable?

YurySolovyov: Or even put this entire method out of component to `lib/player.js`, cause it fits there better IMO.

KeitIG:
Ok, the threshold reached check condition should be in a single method. So we have `isThresholdReached` that check if the action should be called, and move the check in `Player.js` indeed.

So in `Player.js`, we have:
``` jsx
isThresholdReached() {
    return !this.tresholdReached && this.audio.currentTime >= this.audio.duration * this.threshold
}
```

There would be some variable names confusions though, you may need to choose more appropriate function or variable name to make it clear.

YurySolovyov:
> ```js
if(!this.tresholdReached && this.audio.currentTime >= this.audio.duration * this.threshold) {
    this.thresholdReached = true;
    return true;
}
```

Side effect in method that should return boolean value does not seem like a good thing.

**Comment by KeitIG** [comment]
What do you suggest ?

**Comment by YurySolovyov** [comment]
Well, let's at least separate them and move out of components.

**Comment by cfollet** [comment]
Maybe the attribute ` this.thresholdReached` can be` this.thresholdReachedFlag`

**Comment by YurySolovyov** [comment]
@cfollet I'd say we should mention that this is track duration threshold.

**Comment by cfollet** [comment]
@YurySolovyov `thresholdReached` to `trackDurationThreshold` ??

**Comment by YurySolovyov** [comment]
Sounds good.

**Comment by cfollet** [comment]
@YurySolovyov Calling it trackDurationThreshold  means this variable **is** a threshold. In fact it is not.
It's a boolean value that flag when the threshold is reached. 
The name `trackDurationThreshold ` would fit for the value `this.audio.duration * this.threshold`. But we are a little pernickety here :)

**Comment by YurySolovyov** [comment]
just drop `track` part, `durationTreshholdReached`. It is player, so it is mostly about tracks anyway :)

**Comment by KeitIG** [comment]
side note:

`Player.js` is meanly meant for audio manipulation, it is the interface between the audio system (`Audio` object for now) and the app.

This means we'll have to get rid of all `Player.getAudio()` one day to make things cleaner.

**Comment by cfollet** [comment]
it is about AudioContext #128 ?

**Comment by YurySolovyov** [comment]
@cfollet I think it is about moving code of this PR out of component(s) to the audio-related place.

**Comment by KeitIG** [comment]
@cfollet Yury is right. It was just to inform you about what the `Player` class is meant to be :) but this does not change about the changes required for this PR ;)

**Comment by cfollet** [comment]
I updated the PR

**Comment by KeitIG** [comment]
minor typos: it's 'Threshold', and not 'Threshhold' ;)

**Comment by cfollet** [comment]
I think it's ok now

**Comment by KeitIG** [comment]
My only concern was what if `state.repeat === 'one'`, but it seems like the threshold is reset anyway when `AppActions.player.previous` or `AppActions.player.next`. Seems good to me, any last thoughts @YurySolovyov  ?

**Comment by YurySolovyov** [comment]
Why not put it into [`https://developer.mozilla.org/en-US/docs/Web/Events/timeupdate`](timeupdate) event handler? It would be a bit more efficient and we can extract the code from components entirely.

**Comment by cfollet** [comment]
@YurySolovyov Sounds great.

**Comment by YurySolovyov** [comment]
@KeitIG why do we use custom timeout and not timeupdate event in the first place?

**Comment by KeitIG** [comment]
@YurySolovyov you mean why there is a `this.state.elapsed` ?

@cfollet timeupdate looks great if perfs keep being fine.
If so, add your event there: `AppActions.js:36`.

**Comment by YurySolovyov** [comment]
I mean why do we need 
```js
componentDidMount() {
  this.timer = setInterval(this.tick, 100);
}
```

KeitIG:
For this and components perfs reason.

screenshot from 2016-09-13 11 10 08

Do you have a better idea ?

YurySolovyov:
Ah, got it.
> it could be paused when there's no track playing.

Sounds good. 
Also, maybe increase timeout a bit? Say, 250 ms -> 4 times per second.

KeitIG:
We can, I've chosen 150 because the progress bar was more fluid that way.

YurySolovyov:
Maybe in another PR though.

KeitIG:
Let's not pollute this PR too much indeed :stuck_out_tongue_closed_eyes:

YurySolovyov:
@cfollet This looks almost perfect, but can you please move `Player.getAudio()` calls to local variable? Should've done it earlier myself

KeitIG:
It's ok for me. I've made it that way so we now where this come from. This will have to move elsewhere anyway.

KeitIG:
Thanks for your work @cfollet !

PR #206, opened 2016-09-14
Easy access to an artist/album with a right click on track

KeitIG:
Code explication is commented when needed.


YurySolovyov:
You decided to go with context menu? Wasn't it still in discussion state?

KeitIG:
A direct link is not whishable with the current TrackRow design. It' s too thin. But I don't want to change it now. So let's go that way for now, it's not a lot of code anyway.

PR #218, opened 2016-09-20
Avoid sync fs.realpathSync calls in reducers. Fixes #204.

### LINKED ISSUE #204: Avoid sync fs.realpathSync calls in reducers
While looking into #167, I noticed
```js
folders.forEach((folder) => {
    musicFolders.push(fs.realpathSync(folder));
});
```
I think it would be better to avoid doing it that way.

cfollet:


src/js/actions/LibraryActions.js
```diff
@@ -71,6 +72,10 @@ export default {
             properties: ['openDirectory', 'multiSelections']
         }, (folders) => {
             if(folders !== undefined) {
+
+                folders.forEach((folder, index, array) => {
+                    array[index] = fs.realpathSync(folder);
```
YurySolovyov: why still sync? we can do it async with promises and dispatch event when done

cfollet: Yup, my bad. I had to learn Promise. 👍

KeitIG: @YurySolovyov  I don't see how you can get rid of synchronous realpath  loop check here

YurySolovyov: It is more like:
```js
Promise.map(folders, (folder) => {
  return new Promise((resolve, reject) => {
    fs.realpath(folder, (err, realpath) => {
      if (err) { return reject(err); }
      resolve(realpath);
    });
  })
}).then(resolvedFolders => {
  store.dispatch({
    type : AppConstants.APP_LIBRARY_ADD_FOLDERS,
    folders: resolvedFolders
  });
})
```

I'd also used [Bluebird's promisification features](http://bluebirdjs.com/docs/api/promisification.html) to avoid doing 
```js
return new Promise((resolve, reject) => { ... });
```

cfollet: I see how map is better with concurrency than a single Promise. 
@YurySolovyov So, we should avoid the code you provided and use a Promisify one? Isn't Promisification more appropriate with API/Modules stuffs?

YurySolovyov: > Isn't Promisification more appropriate with API/Modules stuffs?

Yes, this is correct. The thing is that I'm not sure how well it works with `import` syntax.
You can create single promisified function higher in the scope via
```js
const realpathAsync = Promise.promisify(fs.realpath);
```

cfollet: Ok, I used the promisify solution

src/js/actions/LibraryActions.js
```diff
@@ -71,9 +71,14 @@ export default {
             properties: ['openDirectory', 'multiSelections']
         }, (folders) => {
             if(folders !== undefined) {
-                store.dispatch({
-                    type : AppConstants.APP_LIBRARY_ADD_FOLDERS,
-                    folders
+                const realpathAsync = Promise.promisify(require('fs').realpath);
```
YurySolovyov: Can you assign this at the top level scope?
Also, you don't need to require fs, you can just use imported variable. (It is ok to use it after it is imported)

YurySolovyov:
@cfollet @KeitIG this LGTM :+1:

KeitIG:
Let's go, thanks for your work.

PR #227, opened 2016-10-06
Add open music folder option on right click

vprigent:
I was looking for that :)

Downloaded some album twice, wanted to delete some old musics, here is a command to open the folder directly on the music.

Don't hesitate to say it if I made a mistake

### DIFF
--- a/src/main-process/ipc.js
+++ b/src/main-process/ipc.js
@@ -110,7 +110,7 @@ class IpcManager {
                     type: 'separator'
                 },
                 {
-                    label: 'Show file in finder',
+                    label: 'Show in file manager',
                     click: () => {
                         shell.showItemInFolder(data.track.path);
                     }
--- a/src/main-process/ipc.js
+++ b/src/main-process/ipc.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { Menu, ipcMain, powerSaveBlocker, app } = require('electron');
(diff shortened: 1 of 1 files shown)


YurySolovyov:
Looks fine to me, except for wording, "finder" is OS specific, we should use neutral term as "file manager" for example.

KeitIG:
Code looks good to me :+1: 

But let's talk a bit about the feature, I'll speak like an investor to a startup: what problem are you trying to solve with that ? I really don't mean to be rude, but could you simply explain me a use-case ?

vprigent:
I've got my library(ies) and now I see that I have some bad information on a music ( unknown album or else ), I right click on the music, open the folder, and can modify my file. Could be done with a more complete integration on Museek. But it's not planned, and perhaps won't (?)

During many years, I downloaded gigs of music. My folder is quit well organized but it would take too much time for it to be perfect. When I don't like a music, I can right click it, access the folder in half the time it would take using my file manager.
There could also be a direct feature on Museek for deleting a set of file, but not planned.

You see where I want to go ?
I currently have 3373 tracks on Museek. I can't afford to spend a day looking into every sub-folder searching for duplicate tracks/albums. I prefer to listen to music and when I see a duplicate track, I open the folder of one of them and remove it.

KeitIG:
Fair enough indeed. Thanks for your work !

PR #229, opened 2016-10-14
Add precisions for developers on README

vprigent:
Because well, seems stupid only when you know it

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,10 +89,14 @@ If you want to report a bug, first, thanks a lot. To help us, please indicate yo
 
 - Fork and clone
 - Master is usually unstable, checkout to a tag to have a stable state of the app
-- `npm install && npm run dev` then run in a separate terminal `electron .`
-- Enable dev mode in the app in the settings view to show DevTools
 
-Webpack will watch for JSX and SASS changes.
+- Install the latest version of electron by running `npm install -g electron`
+- NPM will install electron globally. You can use it on every of your electron projects now.
+
+- `npm install && npm run dev` then run in a separate terminal `electron .`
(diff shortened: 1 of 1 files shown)


KeitIG:
Well, in this case, could you add a list of steps to set up Electron with Museeks ? (It's true it's not really clear)

vprigent:
This should do the job.

We could perhaps add some links for good practice guides ?
However I have none for electron/node/react

README.md
```diff
@@ -89,10 +89,14 @@ If you want to report a bug, first, thanks a lot. To help us, please indicate yo
 
 - Fork and clone
 - Master is usually unstable, checkout to a tag to have a stable state of the app
-- `npm install && npm run dev`
-- Enable dev mode in the app in the settings view to show DevTools
 
-Webpack will watch for JSX and SASS changes.
+- Install the latest version of electron by running `npm install -g electron`
+- NPM will install electron globally. You can use it on every of your electron projects now.
```
YurySolovyov: > You can use it on every of your electron projects now.

I'd rephrased that. Latest electron might not work for just about any electron project.
Something like "You can use electron globally now with `electron [appdir]` command." or something like that,

KeitIG: I'd put the manual electron install too. I don't like to use global package for that.

My recommended way is to download Electron, then place the app in `resources/app`.

vprigent:
I read what I could about the manual electron. Never used that way before and never would again, it's a lot more painful as a beginner, even if it's safer

KeitIG:
> it's a lot more painful as a beginner, even if it's safer

That's the point.

Otherwise it's fine too me, thanks for that.

PR #235, opened 2016-10-21
Pause player when entering in sleep mode (fix #232)

### LINKED ISSUE #232: Stop playing music when recovering from Sleep
So I've been really enjoying Museeks and am pretty sure I'll be replacing iTunes permanently with it. Here's one thing that caught me off guard though:

Expectation: If I have music playing, when I close my laptop and open it later in the day (or early in the morning, as happened just now) I expect the music to not be playing.

What happens: when I open my computer again the music continues playing, which can be surprising.

For context: iTunes and Spotify stop playing music on closing the laptop. 

System: Macbook Pro w/ OS X El Capitan.

KeitIG:
Fix #232 

Oddly not working on Linux


YurySolovyov:
Maybe try sync version of send for Linux?

KeitIG:
The problem is not the send method, it's just the event is never called. See https://github.com/electron/electron/issues/7699

KeitIG:
Merging this, we'll see about the linux bug later as it looks like an upstream bug from Electron.

PR #236, opened 2016-10-23
Refactor action creators export.

### LINKED ISSUE #222: Refactor action creators exports.
I think we discussed this on Gitter, but it is better to create a separate issue for this.
Basically, idea is that we should not embed methods directly into objects we're exporting, and and only export references to the actual functions.
So that instead of 
```js
export default {
    foo: () => { ... }
};        
```
we do
```js
const foo = () => { ... };
export default {
    foo
};        
```
This will simplify code reuse and refactoring.

cfollet:
Not sure about the AppAction refactor. Need review please.
Fixes #222.


src/js/actions/AppActions.js
```diff
@@ -3,146 +3,145 @@ import app    from '../lib/app';
...
+    // Remember dimensions and positionning
+    const currentWindow = app.browserWindows.main;
+
+    currentWindow.on('resize', () => {
+        saveBounds();
+    });
+
+    currentWindow.on('move', () => {
+        saveBounds();
+    });
```
YurySolovyov: I guess this can be just `currentWindow.on('move', saveBounds);` and so on.

cfollet: Thanks for the review. I will adapt as needed 👍

cfollet: AppAction members are used in  the route file. Like [this](https://github.com/KeitIG/museeks/blob/master/src/js/router/routes.js#L29). How do we handle that?

src/js/actions/AppActions.js
```diff
@@ -3,146 +3,145 @@ import app    from '../lib/app';
...
 import LibraryActions       from './LibraryActions';
 import PlaylistsActions     from './PlaylistsActions';
-import NotificationsActions from './NotificationsActions';
 import PlayerActions        from './PlayerActions';
-import QueueActions         from './QueueActions';
 import SettingsActions      from './SettingsActions';
 
 const globalShortcut = electron.remote.globalShortcut;
 const ipcRenderer    = electron.ipcRenderer;
+const player         = PlayerActions;
```
cfollet: AppAction members are used in  the route file. Like [this](https://github.com/KeitIG/museeks/blob/master/src/js/router/routes.js#L29). How do we handle that?

KeitIG:
@cfollet The last two PR probably create conflicts with your branch, don't forget to be up-to-date with master :)

KeitIG:
@cfollet I undestand the struggle with `AppActions` and `AppActions.app`. just simply translate it, I will rename some actions after this PR is merged.

src/js/actions/PlaylistsActions.js
```diff
@@ -7,111 +7,117 @@ import app from '../lib/app';
...
+        }
+    });
+};
+
+export default{
+    load,
+    refresh,
+    create,
+    rename,
+    delete:remove,
```
YurySolovyov: I'd changed it to remove in all call sites so we can export in nicely.

src/js/actions/PlaylistsActions.js
```diff
@@ -7,111 +7,117 @@ import app from '../lib/app';
...
+                type: AppConstants.APP_PLAYLISTS_REFRESH,
+                playlists
+            });
+        }
+    });
+};
 
-        app.models.Playlist.remove({ _id }, (err) => {
+const create = (name, redirect = false, callback = () => {
+}) => {
```
YurySolovyov: Or rather put it in variable somewhere at the top like:
```js
const noop = () => {};
```
and write 
```js
const create = (name, redirect = false, callback = noop) => { ... }
```

cfollet:
Should be ok now

YurySolovyov:
LGTM. /cc @KeitIG

KeitIG:
Thanks, I'll test the app this afternoon.

KeitIG:
Working fine ! Thanks for this :+1:

YurySolovyov:
Thanks @cfollet !

PR #237, opened 2016-10-24
System tray icon

### LINKED ISSUE #219: System Tray Icon
Can you add an icon in the system tray and allow museeks to run in the background.

KeitIG:
Fix #219


src/js/actions/AppActions.js
```diff
@@ -97,7 +103,16 @@ const AppActions = {
         },
 
         close: function() {
-            app.browserWindows.main.close();
+
+            if(app.config.get('useTray')) {
```
YurySolovyov: Weird key name. How about `minimizeToTray` ?

src/js/actions/SettingsActions.js
```diff
@@ -122,6 +122,16 @@ export default {
         AppActions.app.restart();
     },
 
+    toggleTray: function() {
```
YurySolovyov: I wonder why not use `checked` property instead of toggling.

src/main-process/tray.js
```diff
@@ -0,0 +1,64 @@
...
+    constructor(win, icon) {
+
+        this.tray = null;
+        this.trayIcon = icon;
+        this.win = win;
+    }
+
+    bindEvents() {
+
+        const self = this;
```
YurySolovyov: I think with arrow functions this is unnecessary

src/js/actions/AppActions.js
KeitIG: Fair

src/js/actions/SettingsActions.js
KeitIG: That's a fair concern, it's just the other actions are made that way. I'll open an issue for that.

src/main-process/tray.js
KeitIG: True

KeitIG:
About the checked property, I'm doing it right now, I'll push it in this PR

PR #243, opened 2016-10-27
Fix tray dimensions on Mac and some Linux distributions

### LINKED ISSUE #240: Tray Icon strange behavior ( edit )
On MacOS, Tray Icon is displayed in full size :
Working on master dev mode, using Electron v1.4.3 with OS X El Capitan on iMac 27", the tray icon is used on tray without any resize of the image.

Tray capture

KeitIG:
Fixes #240


YurySolovyov:
I'll take a look this evening

vprigent:
Looks good, tested on Ubuntu with their new tray system.
Why not just resize the base image ? Less calculation, less code

YurySolovyov:
Looks better, but blurry.
image

KeitIG:
@vprigent If I agree the 256 icon is not needed, I'll take the 128 or the 64 at least in case of retina displays.

I'll commit that tomorrow.

YurySolovyov:
@KeitIG yeah, still better than oversized one. :shipit:

PR #250, opened 2016-11-14
Make Museeks a single instance app (fix #246)

### LINKED ISSUE #246: Can't open multiple instances of Museeks
I think it's due to a database lock issue. Do you think it's fine if we can't open multiple instances of Museeks at a time ?

KeitIG:


cfollet:
Looks good to me as it's the same as the [documentation](https://github.com/electron/electron/blob/master/docs/api/app.md#appmakesingleinstancecallback)

KeitIG:
I literally stole it from there haha

PR #252, opened 2016-11-15
Place the main frame in the middle of the screen when offscreen at startup (fix #228)

### LINKED ISSUE #228: Window location incorrect for multiple monitors
When Museeks was last used on a second monitor, it reopens in the same position.
For me, this is {-1312, 90}.
If I now only have one monitor connected Museeks is off the screen to the left.
The only way to get it back is to reconnect the monitor or edit the config file.
If both monitors were active I could use Win+Shift+Right to move it to my current one but this doesn't work when only one is active.
When switching from two screens to one when Museeks is already open, it successfully moves to the active one.

Only noticed on Windows 10, Museeks 0.7.1, can possibly try linux if desperate.

KeitIG:


YurySolovyov:
I think it is better to put things like `(fix #228)` to the body of the issue to make it actually linkable

src/main-process/main.js
```diff
@@ -61,6 +61,21 @@ app.on('ready', () => {
...
+    // check if the browser window is offscreen
+    const display = electron.screen.getDisplayNearestPoint(bounds).workArea;
+
+    const onScreen = bounds.x >= display.x
+        && bounds.x + bounds.width <= display.x + display.width
+        && bounds.y >= display.y
+        && bounds.y + bounds.height <= display.y + display.height;
+
+    if(!onScreen) {
+        delete bounds.x;
```
KeitIG: Do you mean "overwrite with the x and y bounds" or "overwrite the whole `bounds` object" ?

YurySolovyov: > overwrite with the x and y bounds

KeitIG: Well, I'll set them to `null` so, because Electron will auto-center the window then

YurySolovyov: can you at least put it into separate method?

KeitIG: Done. Cleanest way I found to do that for now.

PR #254, opened 2016-11-19
Always show tray, tray>quit quits entire app

### LINKED ISSUE #238: beforeunload handler causes app to close on refresh from DevTools
I tracked it down to this: [bc7b401](https://github.com/KeitIG/museeks/commit/bc7b401fd588c6bd58254fc022230b6ba9d2d749#diff-903dcab71c64a976160e32af8a47868fR77)

``` js
window.addEventListener('beforeunload', (e) => {
    // See http://electron.atom.io/docs/api/browser-window/#event-close
    e.returnValue = false;
    this.app.close();
});
```

We should find a better way to do this.

dkniffin:
Fixes #238

### DIFF
--- a/src/js/actions/AppActions.js
+++ b/src/js/actions/AppActions.js
@@ -63,12 +63,6 @@ const init = () => {
         e.preventDefault();
     }, false);
 
-    window.addEventListener('beforeunload', (e) => {
-        // See http://electron.atom.io/docs/api/browser-window/#event-close
-        e.returnValue = false;
-        close();
-    });
-
     // Remember dimensions and positionning
     const currentWindow = app.browserWindows.main;
 
@@ -89,7 +83,6 @@ const close = () => {
     if(app.config.get('minimizeToTray')) {
 
         app.browserWindows.main.hide();
-        ipcRenderer.send('showTray');
 
     } else {
 
--- a/src/js/actions/AppActions.js
+++ b/src/js/actions/AppActions.js
@@ -63,6 +63,11 @@ const init = () => {
         e.preventDefault();
     }, false);
 
+    window.addEventListener('beforeunload', (e) => {
+        e.returnValue = false;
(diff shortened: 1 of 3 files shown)


dkniffin:
@KeitIG I believe that functionality was already built into the app. Unless I'm misunderstanding, that's how it seems to work now.

KeitIG:
Go in settings -> Interface -> use native frame.

Then when I click on the close native button, the app closes when it should minimize to tray. If you don't know how to do this I can do it if you want.

dkniffin:
Oh! I see. Sorry, I didn't see that setting. I'll see if I can figure this out.

KeitIG:
Great, if you don't know how to deal with it, just ask.

ref: http://electron.atom.io/docs/api/browser-window/#event-close

dkniffin:
@KeitIG Alright, I made some more changes. I _think_ it meets all the requirements you mentioned. The one thing I wish it would do that I couldn't quite figure out is when you right click in the dock, and click "quit", it should be the same as quit from tray, but instead, it doesn't close it.

KeitIG:
Ok, check the errors given by the CI build and I think we're good to go

dkniffin:
Done.

dkniffin:
I also added `bindEvents` back, because I think it could be useful later. I was actually about to use it for the other tray actions.

KeitIG:
Can you fix the build errors ?

see https://travis-ci.org/KeitIG/museeks/builds/177480653

dkniffin:
Yep, sorry about that. I've also installed the atom-eslint plugin, so I should see these errors before commiting.

KeitIG:
You can use npm run lint to check that locally too

dkniffin:
Yea, I just forget to do it 😝

KeitIG:
Let's go.

PR #265, opened 2016-11-21
Fix volume slider behavior

dkniffin:


dkniffin:
I wonder if this logic should be refactored into a redux action and maybe a reducer for the exponential stuff.

KeitIG:
This is not working as expected indeed. I found the bug, and it was not where I thought ! I'm closing this and opening a PR, thanks anyway, sorry for that.

KeitIG:
#266

Issue #263, opened 2016-11-21
Remember values between open/close


dkniffin:
This seems to have been implemented before in #42, but it's either broken or removed now.

I'd like for museeks to remember:

- [x] volume
- [x] window size
- [x] shuffle
- [x] repeat
- [x] ~~currently playing song (maybe)~~

KeitIG:
Currently playing song -> nope.

If the other things do not work, there were a regression somewhere after 0.7.1.

KeitIG:
Volume works, window size too

dkniffin:
Volume does not work for me. Maybe it's because I'm running it via the electron app, instead of standalone museeks?

KeitIG:
nah, my bad, it does not work indeed.

KeitIG:
U_U I think the shuffle and the loop were never implemented

dkniffin:
Yea, I figured as much. 👍

PR #266, opened 2016-11-22
Fix forgotten volume, playback and muted values on startup (fix #256, #263)

### LINKED ISSUE #256: Initial volume doesn't match volume bar
The volume seems to start out at a certain level, but the initial value for volume slider in the app doesn't match up with that value, so as soon as you move the slider, the music gets a lot louder.

KeitIG:


YurySolovyov:
Does exponentiation always results in the max/min value at highest/lowest positions?
I wonder if it makes sense to add some kind of rounding to make 99.9 -> 100 and/or 0.01 -> 0, or whatever ranges we use.

KeitIG:
@YurySolovyov it has a step of 0.01, we can increase it a bit if needed.

YurySolovyov:
What I meant is: does dragging volume control to the full-right makes value become 100% ? or it is 99.9%?

KeitIG:
ah, 100.

The input is in [0, 1] range.

```
Math.pow(0, 4) -> 0;
Math.pow(0.5, 4) -> 0.0625
Math.pow(1, 4) -> 1;
```

src/js/actions/PlayerActions.js
```diff
@@ -55,16 +55,33 @@ const previous = () => {
...
 };
 
 const setVolume = (volume) => {
+
     if(!isNaN(parseFloat(volume)) && isFinite(volume)) {
 
-        Player.setAudioVolume(volume);
+        Player.setAudioVolume(Math.pow(volume, 4));
 
         app.config.set('audioVolume', Math.pow(volume, 4));
```
dkniffin: Since e're doing this twice, we should pull it out to a variable `exponentialVolume`, and use that in both spots.

dkniffin: Also, what do you think of refactoring this method to dispatch a `APP_PLAYER_VOLUME_CHANGE`, and put this logic into a reducer?

src/js/components/Header/VolumeControl.react.js
```diff
@@ -54,7 +54,7 @@ export default class VolumeControl extends Component {
                            min={ 0 }
                            max={ 1 }
                            step={ 0.01 }
-                           defaultValue={ this.state.volume, 1 / 4 }
+                           defaultValue={ Math.pow(this.state.volume, 1 / 4) }
```
dkniffin: I might be wrong, but I believe this should go to line 24, so that line 51 is also updated correctly.

KeitIG: Fair

src/js/actions/PlayerActions.js
KeitIG: > Also, what do you think of refactoring this method to dispatch a APP_PLAYER_VOLUME_CHANGE, and put this logic into a reducer?

No for multiple reasons:
- It wouldn't respect Flux pattern (the action update the config file)
- I think all the exponential stuff should always take place in the component, as it both receives and triggers values.

KeitIG:
What are your questions about ?

src/js/components/Header/VolumeControl.react.js
```diff
@@ -65,7 +65,7 @@ export default class VolumeControl extends Component {
 
     setVolume(e) {
         AppActions.player.setVolume(parseFloat(e.currentTarget.value));
-        this.setState({ volume: e.currentTarget.value });
+        this.setState({ volume: Math.pow(e.currentTarget.value, 4) });
```
dkniffin: Hmm. I think this commit re-breaks exponential volume control. This line will set the state to the correct value, but line 67 is the one that actually changes the audio volume, right?

dkniffin:
Oh, I asked them in the review comments. You already addressed them with your commit.

KeitIG:
I think it "should" be good now. woo.

src/js/components/Header/VolumeControl.react.js
```diff
@@ -32,7 +32,7 @@ export default class VolumeControl extends Component {
     }
 
     getVolumeIcon(volume, muted) {
-        return muted || volume === 0 ? 'volume-off' : volume > 0.5 ? 'volume-up' : 'volume-down';
+        return muted || volume === 0 ? 'volume-off' : Math.pow(volume, 1 / 4) > 0.5 ? 'volume-up' : 'volume-down';
```
dkniffin: Ah, yea. These values are based on the linear values. Good catch.

One more knitpick. You can take it or leave it: since we're now doing this twice, I think it'd be good to pull it out into a function (`LinearizeVolume` or something) to DRY it up. It's probably also a good idea to do that with the `Math.pow` (`ExponentiateVolume` or something), so it's clear why we're doing it.

KeitIG: You don't see any problem if `LinearizeVolume` is in the component file as we're only using it here right ?

dkniffin: Yea, I think that makes sense. We've essentially just changed the component to an exponential slider, and any outside observers can just use the exponential value. They shouldn't have a need for the linear value.

However, I _could_ see another volume slider needing the linear volume value. eg: if we put a slider in the tray. But we can address that concern when we get to it.

KeitIG: Roger that

src/js/components/Header/VolumeControl.react.js
```diff
@@ -21,7 +34,7 @@ export default class VolumeControl extends Component {
 
         this.state = {
             showVolume : false,
-            volume     : Math.pow(audio.volume, 1 / 4),
+            volume     : unsmoothifyVolume(audio.volume, 1 / 4),
```
dkniffin: I think this should be `unsmoothifyVolume(audio.volume, 4)`, since you're doing the `1 /` above

KeitIG: Wrong :p we're both wrong.

DARN.

I'll squash commits.

KeitIG:
Are you able to accept the whole review?

dkniffin:
I think I did accept the whole PR, but I don't count because I don't have write access. I think you need to make me a collaborator

PR #267, opened 2016-11-22
Remove title from tray

dkniffin:
Adding the title makes it look a bit out of place.

Before:

image

After:
image


KeitIG:
I didn't know what the `title` option was about. I thought it was a kind of "hover" thing like the title attribute in HTML. 

Now I know :laughing:

dkniffin:
😄 I probably would have thought the same thing. I think `setToolTip` does that.

PR #272, opened 2016-11-23
Refactor Tracklist

dkniffin:
Mostly, I broke a few really large methods into smaller methods.

I also want to note that [this switch](https://github.com/KeitIG/museeks/blob/master/src/js/components/Shared/TracksList.react.js#L217-L282) only really needs the default case, because the other two are covered by the default.

### DIFF
--- a/src/js/components/Shared/TracksList.react.js
+++ b/src/js/components/Shared/TracksList.react.js
@@ -189,7 +189,7 @@ export default class TracksList extends Component {
                 let playingIndicator = undefined;
 
                 if(trackPlayingId === track._id) {
-                    playingIndicator = <PlayingIndicator state={ Player.getAudio().paused ? 'pause' : 'play' } />;
+                    playingIndicator = <PlayingIndicator state={ this._pausePlayState() } />;
                 }
 
                 return(
@@ -233,6 +233,10 @@ export default class TracksList extends Component {
                 </div>
             );
         });
+    }
+
+    _pausePlayState() {
+        Player.getAudio().paused ? 'pause' : 'play';
     }
 
     _toggleSelectionById(id) {
--- a/src/js/components/Shared/TracksList.react.js
+++ b/src/js/components/Shared/TracksList.react.js
@@ -291,23 +291,19 @@ export default class TracksList extends Component {
         const selected = this.state.selected,
             tracks     = this.props.tracks;
 
-        const length = tracks.length;
-        let i;
-        for(i = 0; i < length; i++) {
-            if(selected.includes(tracks[i]._id)) break;
-        }
+        const firstSelectedTrackIdx = tracks.findIndex((track) => selected.includes(track._id));
 
         switch(e.keyCode) {
             case 38: // up
-                this._onUp(i, tracks);
+                this._onUp(firstSelectedTrackIdx, tracks);
                 break;
 
             case 40: // down
-                this._onDown(i, tracks);
+                this._onDown(firstSelectedTrackIdx, tracks);
                 break;
 
             case 13: // enter
-                this._onEnter(i, tracks);
+                this._onEnter(firstSelectedTrackIdx, tracks);
                 break;
         }
     }
--- a/src/js/components/Shared/TracksList.react.js
+++ b/src/js/components/Shared/TracksList.react.js
@@ -140,7 +140,9 @@ export default class TracksList extends Component {
         const selected = this.state.selected,
             tracks     = this.props.tracks;
 
-        const firstSelectedTrackIdx = tracks.findIndex((track) => selected.includes(track._id));
+        const firstSelectedTrackIdx = tracks.findIndex((track) => {
+            selected.includes(track._id);
+        });
 
         switch(e.keyCode) {
             case 38: // up
@@ -157,16 +159,16 @@ export default class TracksList extends Component {
         }
     }
 
-    _isReftClick(e) {
-        e.button === 0;
+    _isLeftClick(e) {
+        return e.button === 0;
     }
 
     _isRightClick(e) {
-        e.button === 2;
+        return e.button === 2;
     }
 
     _isSelectableTrack(id) {
-        !this.state.selected.includes(id);
+        return !this.state.selected.includes(id);
     }
 
     _buildTrackTiles() {
@@ -273,14 +275,8 @@ export default class TracksList extends Component {
             case 1: {
 
                 const onlySelected = selected[0];
-                let onlySelectedIndex;
 
-                for(let i = 0, length = tracks.length; i < length; i++) { // STH wrong here
-                    if(tracks[i]._id === onlySelected) {
-                        onlySelectedIndex = i;
-                        break;
-                    }
-                }
+                const onlySelectedIndex = tracks.findIndex((track) => track._id === onlySelected);
 
                 if(index < onlySelectedIndex) {
                     for(let i = 1; i <= Math.abs(index - onlySelectedIndex); i++) {
--- a/src/js/components/Shared/TracksList.react.js
+++ b/src/js/components/Shared/TracksList.react.js
@@ -288,57 +288,58 @@ export default class TracksList extends Component {
     }
 
     onKey(e) {
-
         const selected = this.state.selected,
             tracks     = this.props.tracks;
 
-        let i = 0;
+        const length = tracks.length;
+        let i;
+        for(i = 0; i < length; i++) {
+            if(selected.includes(tracks[i]._id)) break;
+        }
 
-        switch(e.keyCode) { // Todo: fix when node not in dom
+        switch(e.keyCode) {
             case 38: // up
-
-                for(let length = tracks.length; i < length; i ++) {
-                    if(selected.includes(tracks[i]._id)) break;
-                }
-
-                if(i - 1 >= 0) {
-                    this.setState({ selected : tracks[i - 1]._id }, () => {
-
-                        const container = document.querySelector('.tracks-list-container .tracks-list-body');
-                        const nodeOffsetTop = (i - 1) * this.rowHeight;
-
-                        if(container.scrollTop > nodeOffsetTop) container.scrollTop = nodeOffsetTop;
-                    });
-                }
+                this._onUp(i, tracks);
                 break;
 
             case 40: // down
-
-                for(let length = tracks.length; i < length; i ++) {
-                    if(selected.includes(tracks[i]._id)) break;
-                }
-
-                if(i + 1 < tracks.length) {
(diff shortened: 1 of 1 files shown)


src/js/components/Shared/TracksList.react.js
```diff
@@ -189,158 +124,224 @@ export default class TracksList extends Component {
     }
 
     selectTrack(e, id, index) {
+        if(e.button === 0 || (e.button === 2 && !this.state.selected.includes(id))) {
```
YurySolovyov: can we move this to some variable or short function? to make it read like 
```js
if (isSlectableTrackOrSomething) {
  // ...
}
```

src/js/components/Shared/TracksList.react.js
```diff
@@ -189,158 +124,224 @@ export default class TracksList extends Component {
...
+                this._onEnter(firstSelectedTrackIdx, tracks);
+                break;
+        }
+    }
 
-                selected = utils.simpleSort(selected, 'asc');
-                this.setState({ selected });
+    _buildTrackTiles() {
+        const self         = this,
+            selected       = this.state.selected,
```
YurySolovyov: can you use 1 `const` per assignment?

src/js/components/Shared/TracksList.react.js
```diff
@@ -189,158 +124,224 @@ export default class TracksList extends Component {
...
-                        break;
-                    }
+        return tracksChunked.splice(tilesScrolled, tilesToDisplay).map((tracksChunk, indexChunk) => {
+
+            const list = tracksChunk.map((track, index) => {
+
+                let playingIndicator = undefined;
+
+                if(trackPlayingId === track._id) {
+                    playingIndicator = <PlayingIndicator state={ Player.getAudio().paused ? 'pause' : 'play' } />;
```
YurySolovyov: can we move it to a separate method? cause now it has if+ternary, that makes it hard to read

src/js/components/Shared/TracksList.react.js
```diff
@@ -189,158 +124,224 @@ export default class TracksList extends Component {
...
-                this.setState({ selected });
-            }
-        }
+
+                return(
+                    <TrackRow
+                        selected={ selected.includes(track._id) }
+                        trackId={ track._id }
+                        key={ index }
+                        index={ (tilesScrolled + indexChunk) * chunkLength + index }
```
YurySolovyov: I'd moved that one to variable too

src/js/components/Shared/TracksList.react.js
```diff
@@ -189,158 +124,224 @@ export default class TracksList extends Component {
...
+                        </div>
+                        <div className='cell cell-genre'>
+                            { track.genre.join(', ') }
+                        </div>
+                    </TrackRow>
+                );
+            });
+
+            const tracksListTileStyles = {
+                transform: `translate3d(0, ${(((tilesScrolled * this.rowHeight * chunkLength) + (indexChunk * this.rowHeight * chunkLength)))}px, 0)`
```
YurySolovyov: This one deserves a variable for sure

src/js/components/Shared/TracksList.react.js
KeitIG: We can't define fonctions/variables for all boolean cast we do...

src/js/components/Shared/TracksList.react.js
```diff
@@ -189,158 +124,224 @@ export default class TracksList extends Component {
...
+            case 1: {
 
-                        const container = document.querySelector('.tracks-list-container .tracks-list-body');
-                        const nodeOffsetTop = (i - 1) * this.rowHeight;
+                const onlySelected = selected[0];
+                let onlySelectedIndex;
 
-                        if(container.scrollTop > nodeOffsetTop) container.scrollTop = nodeOffsetTop;
-                    });
+                for(let i = 0, length = tracks.length; i < length; i++) { // STH wrong here
```
YurySolovyov: [`Array.prototype.findIndex()`](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex)

src/js/components/Shared/TracksList.react.js
```diff
@@ -189,158 +124,224 @@ export default class TracksList extends Component {
...
                 }
-                break;
 
-            case 40: // down
+                if(index < onlySelectedIndex) {
+                    for(let i = 1; i <= Math.abs(index - onlySelectedIndex); i++) {
+                        selected.push(tracks[onlySelectedIndex - i]._id);
+                    }
+                } else if(index > onlySelectedIndex) {
+                    for(let i = 1; i <= Math.abs(index - onlySelectedIndex); i++) {
```
YurySolovyov: I wonder if we can make it more functional meaning to use something like `.map` or `.reduce`

src/js/components/Shared/TracksList.react.js
YurySolovyov: we can do for non-trivial ones

src/js/components/Shared/TracksList.react.js
YurySolovyov: if @KeitIG is ok with that I'd also enabled eslint rule for that

src/js/components/Shared/TracksList.react.js
dkniffin: Yea, I like this idea. It makes it a little easier to understand what's going on at a glance. I'll do it. 👍

dkniffin:
Thanks for your feedback, @YurySolovyov. I like your suggestions. I'll update the PR in a little bit.

src/js/components/Shared/TracksList.react.js
dkniffin: Actually, I don't think that works here.

src/js/components/Shared/TracksList.react.js
dkniffin: Yea, we probably could...I'm more inclined to delete all of `case 0` and `case 1`, because they're not necessary.

src/js/components/Shared/TracksList.react.js
```diff
@@ -189,158 +124,244 @@ export default class TracksList extends Component {
...
-                }
+            case 13: // enter
+                this._onEnter(firstSelectedTrackIdx, tracks);
+                break;
+        }
+    }
 
-                selected = utils.simpleSort(selected, 'asc');
-                this.setState({ selected });
+    _isReftClick(e) {
```
KeitIG: REFT REFT REFT

KeitIG:
```
_isSelectableTrack(id) {
    !this.state.selected.includes(id);
}
```

I'm sorry, but why aren't there any return statement ?

dkniffin:
@KeitIG With es6, you don't have to add `return` to functions anymore. It's implied that the last line in the function is meant to be returned. But, since we're using `return` everywhere else, I'll go ahead and add it to these.

src/js/components/Shared/TracksList.react.js
dkniffin: Oh, yep. You're right. I was trying to do something different.

src/js/components/Shared/TracksList.react.js
```diff
@@ -189,158 +124,247 @@ export default class TracksList extends Component {
...
+                const onlySelected = selected[0];
+                let onlySelectedIndex;
+
+                // for(let i = 0, length = tracks.length; i < length; i++) { // STH wrong here
+                //     if(tracks[i]._id === onlySelected) {
+                //         onlySelectedIndex = i;
+                //         break;
+                //     }
+                // }
+                const onlySelectedIndex = tracks.findIndex((track) => return track._id === onlySelected);
```
KeitIG: Use  `bla.map((elem) => elem.thing)` for return statements in fat arrow function, but not in plain function with braces (like `isRightClick: function() {}`)

src/js/components/Shared/TracksList.react.js
```diff
@@ -189,158 +124,240 @@ export default class TracksList extends Component {
...
+        return tracksChunked.splice(tilesScrolled, tilesToDisplay).map((tracksChunk, indexChunk) => {
+
+            const list = tracksChunk.map((track, index) => {
+
+                const trackRowIndex = (tilesScrolled + indexChunk) * chunkLength + index;
+
+                let playingIndicator = undefined;
+
+                if(trackPlayingId === track._id) {
+                    playingIndicator = <PlayingIndicator state={ this._pausePlayState() } />;
```
YurySolovyov: Not sure if @KeitIG would agree, but I'd let all rows have an indicator and encapsulated rendering playing state into them, so you don't have to do conditions here

src/js/components/Shared/TracksList.react.js
YurySolovyov: If you're sure about that, I'm all into removing the code :)

src/js/components/Shared/TracksList.react.js
dkniffin: I don't understand what you mean here. Could you provide an example, or explain it in other words?

src/js/components/Shared/TracksList.react.js
dkniffin: Yea, I mentioned it to @KeitIG in gitter. The default case covers the other two, so we don't really need the other two cases. The _only_ reason I can think of to keep them is a performance reason, but it seems perfectly performant to me without them.

src/js/components/Shared/TracksList.react.js
YurySolovyov: Instead of deciding about rendering indicator at `TracksList` level, it might be better to decide it on
`TracksRow` level.

Example:
```js
const isCurrent = trackPlayingId === track._id;
const playingIndicator = <PlayingIndicator 
                            state={ this._pausePlayState() } 
                            isCurrent = { isCurrent } 
                         />;

// in TrackRow.react.js:
render() {
    if (!this.props.isCurrent) { return null; } // will render nothing

    // ... render if it IS current, just as it is now
}
```

src/js/components/Shared/TracksList.react.js
YurySolovyov: Then drop them.

src/js/components/Shared/TracksList.react.js
dkniffin: Ah, gotcha. 👍

YurySolovyov:
I'm ok with landing this. We can take care of linting later.

dkniffin:
I'm gonna make a few more changes before merging:

- [ ] converting `_methods` to `methods`
- [ ] Move playing indicator into TracksRow
- [ ] Delete two unused cases

KeitIG:
Nice job !

Issue #264, opened 2016-11-21
Tray: show current track metadatas


KeitIG:

dkniffin:
This was done via #274.

PR #274, opened 2016-11-25
Display track metadata in tray

dkniffin:
#264 This was harder than I thought, but I finally got it. 😌 
screen shot 2016-11-25 at 5 45 57 pm

### DIFF
--- a/src/main-process/tray.js
+++ b/src/main-process/tray.js
@@ -17,6 +17,9 @@ class IpcManager {
             {
                 label: 'Not playing',
                 enabled: false
+            },
+            {
+                type: 'separator'
             }
         ];
 
@@ -97,7 +100,7 @@ class IpcManager {
                     break;
                 case 'trackStart': {
                     const trackMetadata = param1;
-                    this.songDetails[0].label = this._trayString(trackMetadata);
+                    this._updateTrayMetadata(trackMetadata);
                     this._setContextMenu('play');
                     break;
                 }
@@ -139,8 +142,27 @@ class IpcManager {
     }
 
 
-    _trayString(metadata) {
-        return `${metadata.title} - ${metadata.artist} - ${metadata.album}`;
+    _updateTrayMetadata(metadata) {
+        this.songDetails = [
+            {
+                label: `Song: ${metadata.title}`,
+                enabled: false
+            },
+            {
+                label: `Artist: ${metadata.artist}`,
+                enabled: false
+            },
+            {
+                label: `Album: ${metadata.album}`,
+                enabled: false
+            },
+            {
+                type: 'separator'
+            }
+        ];
+        // this.songDetails[0].label = this._trayString(metadata);
+
+        // return `${metadata.title} - ${metadata.artist} - ${metadata.album}`;
     }
 
 }
--- a/src/main-process/tray.js
+++ b/src/main-process/tray.js
@@ -13,6 +13,12 @@ class IpcManager {
         this.win = win;
         this.contextMenu;
 
+        this.songDetails = [
+            {
+                label: 'Not playing',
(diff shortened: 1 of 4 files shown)


src/main-process/tray.js
```diff
@@ -113,13 +128,43 @@ class IpcManager {
...
+                label: `Album: ${metadata.album}`,
+                enabled: false
+            },
+            {
+                type: 'separator'
+            }
+        ];
+        // this.songDetails[0].label = this._trayString(metadata);
+
+        // return `${metadata.title} - ${metadata.artist} - ${metadata.album}`;
```
YurySolovyov: Seems like some leftovers

YurySolovyov:
I personally don't get why people use `_foo` for method names, if you want to make it private, just put it in closure and don't export it, or just drop `_` from the name.

src/main-process/tray.js
dkniffin: Yep. At first, I tried putting all the metadata on one line, but it didn't look very good. I'll remove it

src/main-process/tray.js
```diff
@@ -113,13 +128,43 @@ class IpcManager {
...
     }
 
     hide() {
 
         this.tray.destroy();
     }
+
+    _setContextMenu(state) {
+        const playPauseItem = state === 'pause' ? this.pauseToggle : this.playToggle;
+        this.tray.setContextMenu(Menu.buildFromTemplate([...this.songDetails, ...playPauseItem, ...this.menu]));
```
dkniffin: 👍

dkniffin:
I'm not sure what you mean by a closure. That was my attempt at doing pseudo-private functions, since javascript doesn't actually support real private functions yet. I'd be fine switching it to do something else.

YurySolovyov:
@dkniffin by closure I mean this:
```js
const privateFn = function(param1, param2) {
  // ...
};

const publicFn = function(param) {
  // some code ...
  // use privateFn
  return privateFn(param, 42);
};


export default {
  publicAPI: publicFn
};
```

This way you don't have to use `_` prefix to make `privateFn` "private".
If you need to use `this`, use `.bind` method, it is pretty fast in V8

dkniffin:
How would that look for a `class`?

YurySolovyov:
For `class` that would look pretty much the same, the downside is that this function won't be a part of class.
I'd not worry about making these private and just given them names that make sense, there is no such thing as "public" API in museeks yet, so I don't think it matters.

KeitIG:
Told you it would not be as easy as it seems :p 

Just about the look, would it be possible to see what this would look like? :

```
[title]
on [album]
by [artist]
```

dkniffin:
I like keeping the artist second, personally:

screen shot 2016-11-26 at 4 44 35 pm

Also, this is something we might be able to have as a user preference, maybe. I'm a big fan of lots of configuration options in apps. Maybe a checkbox for what's shown, so if people don't care about what album, they can disable that, etc. What do you guys think?

YurySolovyov:
1. Can we center labels?
2. What happened to idea of small window for mini-player?

YurySolovyov:
> I'm a big fan of lots of configuration options in apps

I'm a big fan of sane defaults. For museeks, if we can't do it right, should we do it at all?
IMO Minimalism is one if the strongest sides of museeks and I'd keep following it.

KeitIG:
Centering label would not look good.

I agree with Yury for minimalization.

About mini-player: later maybe, it's a longer stuff to code

dkniffin:
Alright, sounds good on all of that. I don't know if we can center the labels, even if we wanted to.

KeitIG:
Ok, good to me, thanks !

PR #283, opened 2016-12-02
Remove onbeforeunload handler

### LINKED ISSUE #238: beforeunload handler causes app to close on refresh from DevTools

KeitIG:
Fix #238 

Tell me if I'm wrong but we don't need it anymore as we always show the tray (poke @dkniffin)

### DIFF
--- a/src/js/actions/AppActions.js
+++ b/src/js/actions/AppActions.js
@@ -80,11 +80,6 @@ const init = () => {
         e.preventDefault();
     }, false);
 
-    window.addEventListener('beforeunload', (e) => {
-        e.returnValue = false;
-        minimize();
-    });
-
     // Remember dimensions and positionning
     const currentWindow = app.browserWindows.main;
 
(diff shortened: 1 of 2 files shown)


dkniffin:
Hmm. No, this breaks it, on OSX at least. If I click X on the native frame, with this change enabled, it closes the window (and stops playing music), and if I try and click show again, I get this error:

image

KeitIG:
Oh right, the native frame.

KeitIG:
@dkniffin This one should work.

PR #287, opened 2016-12-02
Fix #286

### LINKED ISSUE #286: Notification display delay
Between the time I click play and the the time the notification is displayed, there's about a 3 second pause. I'd like to remove that. I'm pretty sure it's either caused by [this line](https://github.com/dkniffin/museeks/blob/master/src/js/actions/AppActions.js#L39) or [this line](https://github.com/dkniffin/museeks/blob/master/src/js/actions/AppActions.js#L42)

dkniffin:


dkniffin:
I could do that, but I think we want it like it is, because the `'playerAction', 'trackStart'` is being used to update metadata in the tray too.

KeitIG:
Indeed. I'm concerned with that long db call, but let's go for the moment.

YurySolovyov:
@dkniffin mind editing PR title to say what was done?

PR #290, opened 2016-12-02
Fix proptypes validation across the app

### LINKED ISSUE #279: Fix React PropTypes validations
I think we missed that one cause we were running in prod mode all the time, but #276 revealed some issues we have

KeitIG:
fix #279

I removed the required option because the proptypes were checked before receiving the props. Yes that' super weird, I don't know why.


YurySolovyov:
LGTM. Good stuff.

Issue #292, opened 2016-12-05
TracksList keyboard shortcuts are broken


KeitIG:
- up key
- down key
- enter key

I probably broke when refactoring TracksList component

dkniffin:
Sorry, I've been distracted the past week. I'll take care of this now.

PR #302, opened 2016-12-14
Fix up, down, and enter keys in tracklist

dkniffin:
#292

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,6 +19,7 @@
     },
     "rules": {
         "array-bracket-spacing": ["error", "never"],
+        "array-callback-return": "error",
         "arrow-parens": ["error", "always"],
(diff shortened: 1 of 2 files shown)


YurySolovyov:
Heh, are there any eslint rules for that?

KeitIG:
I guess not. Eslint wouldn't know if it's a static method or not.

dkniffin:
It wouldn't help with this problem, but there's [this](http://eslint.org/docs/rules/consistent-return)

YurySolovyov:
maybe this one http://eslint.org/docs/rules/array-callback-return ?

dkniffin:
Good find, @YurySolovyov! That caught it. I added the config for it.

KeitIG:
Thanks for the fix.

Issue #41, opened 2015-10-29
Add minimize/maximize window controls


KeitIG:
without destroying design, that'd be great.

KeitIG:
Waiting for https://github.com/FortAwesome/Font-Awesome/issues/1582

WaynePhipps:
https://github.com/FortAwesome/Font-Awesome/issues/1582#issuecomment-21042857

Hi guys, 

I don't understand the comment

> Waiting for FortAwesome/Font-Awesome#1582

Wondering what you're waiting for here? Is there a metric or milestone involved?

The issue dates back to July 2013 and seems to have lots of support. Just wondering on the likelihood of this being completed any-time in the next 6 months?

Is there some help the community can provide?

YurySolovyov:
[`octicons`](https://octicons.github.com/) seem to have:
- [`.octicon-dash`](https://octicons.github.com/icon/dash/) for minimize
- [`.octicon-link-external`](https://octicons.github.com/icon/link-external/) for maxmize
- [`.octicon-x`](https://octicons.github.com/icon/x/) for close

Though this would add package (does not seem to be a big deal) + need to figure out how to add only what we use instead of using the whole package (some webpack tweaks?).
Might also need to do some experiments to make sure it looks consistent enough.

KeitIG:
quick idea: make an option to have electron `frame` at true. There would be just some CSS tweaks to do (borders to remove)

KeitIG:
https://materialdesignicons.com/

KeitIG:
Things are being unlocked and these icons will be released in FA 4.7.0

KeitIG:
On the road :]

KeitIG:
FA 4.7.0 is out, I'll do that soon

YurySolovyov:
@KeitIG I tried it, had some issues with empty cover icon, I thought you might want to know.

dkniffin:
Alright, got it in there: 
![image](https://cloud.githubusercontent.com/assets/1680080/21945775/36cc1e58-d9ab-11e6-8156-13d9f4064135.png)

I'll make a PR

Issue #135, opened 2016-08-09
Better library updates


KeitIG:
## What

The current problem is when refreshing the library, everything is reset, including playlists.

## How

When a file is in DB, it should not be removed with a refresh. The refresh option should not exist anymore, or should be more explicit and with a warning telling you that everything will be reset.
- [x] Add files or olders
- [x] Reset library should be more explicit (warning or danger color)
- [x] Don't freeze the interface when refreshing library
- [x] Be able to add more folders/tracks when the library is already refreshing
- [x] Add files to library by drag & dropping them on the main BrowserWindow or custom dropZone
    - [ ] Directly add files to playlist if on playlist view
- [x] Remove tracks from `TracksList` #295 
- [x] Refresh progress if footer

YurySolovyov:
> Is "reset library" still needed ?

IMO this depends on how smart library refresh can be or how smart we want it to be.
If it is ok to wipe out current library and start over, then yeah, we can merge these actions.
One point to consider is that after hard reset users will probably want to do refresh anyway.

Building smarter refresh has some challenges - e.g. how do you handle file renames? What files are considered "same" or "equal" or "up-to-date" ... stuff like that.

KeitIG:
The goal is to add tracks to the library when adding folders, and remove them when removing a folder. No file rename handle or whatever.

My only concern is if a track is in a playlist. Currently, a playlist looks like that:
```
// playlist
{
    name,
    type   : 'playlist',
    tracks :  [] // array of _ids
};
```

I'm ok with letting unused _ids in playlists, as they won't be fetched by NeDB anyway, and it's hard to look for them.

OR

change data structure and do sth like: 
```
{
    _id,
    name,
    type   : 'playlist',
}
```

and another collections to join them: 
```
{
    trackId
    playlistId,
    order
}
```

which looks a bit more like SQL

KeitIG:
This is not a problem with an `order` key and/or an unique _id

YurySolovyov:
> The goal is to add tracks to the library when adding folders

This makes a lot of sense, we can also put refresh buttons on individual folder items, and drop refresh/reset buttons completely

YurySolovyov:
An idea: maybe as part of this effort, let's move library management view to a separate tab at the bottom? I don't really feel it belongs to settings...

dkniffin:
[...]

PR #307, opened 2017-01-13
Add min/max window controls

dkniffin:
#41

image

### DIFF
--- a/src/js/components/Header/WindowControls.react.js
+++ b/src/js/components/Header/WindowControls.react.js
@@ -22,12 +22,22 @@ export default class WindowControls extends PureComponent {
 
         return (
             <div className='window-controls'>
-                <button className='window-control' onClick={ this.winClose }>&times;</button>
+                <button className='window-control fa fa-window-close' onClick={ this.winClose } />
+                <button className='window-control fa fa-window-minimize' onClick={ this.winMinimize } />
+                <button className='window-control fa fa-window-maximize' onClick={ this.winMaximize } />
             </div>
         );
     }
 
     winClose() {
         AppActions.close();
+    }
+
+    winMinimize() {
+        AppActions.minimize();
+    }
+
+    winMaximize() {
+        AppActions.maximize();
     }
 }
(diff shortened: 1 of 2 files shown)


YurySolovyov:
How about screenshot with native frame?

dkniffin:
With the native frame, it doesn't show up at all.

image

YurySolovyov:
LGTM, thought you need to wait for @KeitIG's :+1:

KeitIG:
I'll review it asap

dkniffin:
@KeitIG I'm fine with the width, font-size and height changes, but I think the `active` state should stay. It feels weird to have a button that doesn't react when pressed, and it doesn't match the behavior of other windows.

YurySolovyov:
I like how atom does it
Dark:
image
Light:
image

KeitIG:
@dkniffin Ye, I should have precised it was because this active state in particular was not relevant/sexy anymore. I'll add one later.

@YurySolovyov If you find sexier icons, don't hesitate to poke me as the one from FA are not really great.

YurySolovyov:
Ones [you mentioned here](https://github.com/KeitIG/museeks/issues/41#issuecomment-240143430) were looking nice, not much else I can think of.
Maybe we'll be able to extract just minimum required svg code if we choose to use them

KeitIG:
I edited my review with some more things to check

YurySolovyov:
I just found that while initial partial screens look ok, when looking at full-sized one, current it looks weird and oversized.

dkniffin:
I'll finish this up soon. I tried re-importing my library and ran into #135, and haven't been using Museeks much in the past week or so. But @KeitIG pointed me at the branch with the fix, so I should be able to do this now.

KeitIG:
@dkniffin bump :)

dkniffin:
@KeitIG Hey. I don't remember where I left off on this. I probably won't have time soon to finish it out, so would you mind finishing it?

KeitIG:
should be in my power

KeitIG:
Closing this, will re-open a new one.

PR #315, opened 2017-02-20
New library management

### LINKED ISSUE #135: Better library updates

### LINKED ISSUE #295: [Feature Suggestion] Remove Tracks from Library/Filesystem
Maybe its already implemented, since I use the latest stable and not the git version.
Would it be possible to add a remove from library option, maybe with a dialog box that pops up asking if the file(s) should be deleted too.
I'd love to have this, listening a bit through my library right now and atm I have to open the explorer for every track I don't want anymore so that i can remove it.
The look from this app is nice and sleek, keep up that awesome work 👍

KeitIG:
fixes #135 
fixes #295

~~**This is an unstable branch with work being in progress, it does not need any review for now**~~


YurySolovyov:
> This is an unstable branch with work being in progress, it does not need any review for now

So, are ANY comments forbidden? :laughing:

KeitIG:
The only purpose of this PR is to force me to finish it haha. Just psychological stuff here.

KeitIG:
Ok @YurySolovyov, I'd like your thoughts on this. 

The only "important" things here, is to be able to import folders when the importer is already running. I already have some ideas how to do that, but I would like to see your ideas to see If I'm wrong or missing something. (basically, do you have any experiences with queues and promises?) Is there a magic-library or do we have to do everything ourselves?

src/js/actions/LibraryActions.js
```diff
@@ -112,22 +65,48 @@ const refresh = () => {
...
+
     const fsConcurrency = 32;
 
-    // Start the big thing
-    app.models.Track.removeAsync({}, { multi: true }).then(() => {
+    let rootFiles; // HACK Kind of hack, looking for a better solution
+
+    // Scan folders and add files to library
+    new Promise((resolve) => {
+        const paths = Promise.map(pathsToScan, (path) => {
```
YurySolovyov: I'd write this that way:
```js
return Promise.map(pathsToScan, (path) => {
    return Promise.props({
      path,
      stat: statAsync(path) 
    });
});
```

src/js/actions/LibraryActions.js
```diff
@@ -141,21 +120,87 @@ const refresh = () => {
...
+
+const refreshProgress = (processed, total) => {
+    store.dispatch({
+        type : AppConstants.APP_LIBRARY_REFRESH_PROGRESS,
+        processed,
+        total,
+    });
+};
+
+const removeFromLibrary = async (tracksIds) => {
```
YurySolovyov: I wonder if it actually works as expected

src/js/actions/LibraryActions.js
```diff
@@ -141,21 +120,87 @@ const refresh = () => {
...
+    dialog.showMessageBox(app.browserWindows.main, {
+        buttons: [
+            'Cancel',
+            'Remove'
+        ],
+        title: 'Remove tracks from library?',
+        message: `Are you sure you want to remove ${tracksIds.length} element(s) from your library?`,
+        type: 'warning'
+    }, (result) => {
+        if(result === 1) {
```
YurySolovyov: I'd declare this constant somewhere in top scope so we can compare to a named constant instead of magic numbers.

src/js/actions/LibraryActions.js
```diff
@@ -141,21 +120,87 @@ const refresh = () => {
...
+            // Remove tracks from the Track collection
+            app.models.Track.removeAsync({ _id: { $in: tracksIds } }, { multi: true }, () => {
```
YurySolovyov: you are using an `*Async` version and still passing a callback?

src/js/components/Settings/SettingsLibrary.react.js
```diff
@@ -18,66 +18,59 @@ export default class SettingsLibrary extends Component {
...
     }
 
-    refreshLibrary() {
-        AppActions.player.stop();
-        AppActions.library.refresh();
+    onDrop(e) {
+        const files = [];
+        const eventFiles = e.dataTransfer.files;
+
+        for(let i = 0; i < eventFiles.length; i++) {
```
YurySolovyov: `e.dataTransfer.files` is not a plain array? if it is, just use regular map

src/js/components/Shared/Dropzone.react.js
```diff
@@ -0,0 +1,70 @@
...
+        this.onDrop = this.onDrop.bind(this);
+        this.onDragEnter = this.onDragEnter.bind(this);
+        this.onDragLeave = this.onDragLeave.bind(this);
+        this.onDragOver = this.onDragOver.bind(this);
+    }
+
+    render() {
+        return (
+            <div
+                className={ `dropzone ${this.state.state}` }
```
YurySolovyov: use `classNames` here

YurySolovyov:
> Is there a magic-library or do we have to do everything ourselves?

I don't think there is a lib that would do everything for us, and IMO bluebird is already pretty good.

The thing here would be to define and implement updating with clearly separated steps.

Pseudo impl:
```js
try {
  const pathsToScan = await getPathsToScan(...);
  const mediaFileCandidates = await getFilesByPath(pathsToScan, ...);
  const flatFilesArray = await flattenArrays(mediaFileCandidates, ...);
  const supportedMediaFiles = filterSupported(flatFilesArray, ...);
  doSomethingWithSupportedFiles();
} catch (e) {
  // handle e
} 
```

src/js/components/Settings/SettingsLibrary.react.js
KeitIG: nope, `FilesList` here, cannot map through it.

YurySolovyov: How about hacking it like `[...e.dataTransfer.files]` ?

KeitIG:
> Not ready for the prime time :)

Well of course it isn't haha.

> The thing here would be to define and implement updating with clearly separated steps.

Nope it's not the thing here.

Imagine I add the first half of my library in Museeks, the scan begins.
Now, during the scan, I want to add the other folders of music. 
We do not want a new process to be created, we should "merge" the file queue that is being processed, and the new files list.

src/js/components/Settings/SettingsLibrary.react.js
KeitIG: u naughty boy

YurySolovyov:
I'm not sure what you mean by a separate process, it would be just another closure with its own context and everything.

The point about merging is valid though, needs a bit more thinking, but I'm sure it is possible to achieve what I described with simpler approach.

src/js/components/Settings/SettingsLibrary.react.js
YurySolovyov: This is the first time I see the benefit of ES6 iteration protocol :rofl: 
Still makes me wonder why it is not a plain array.

KeitIG:
> I'm not sure what you mean by a separate process, it would be just another closure with its own context and everything.

We could typically fall in another ERR_FILE_OPEN error because we're opening too many files at the same time.

I'll think a bit about a way to achieve this. This cannot be merged as long as It's not solved

YurySolovyov:
**One possible way** to do it would be to have a queue based on plain promises and enqueuing new work with `.then`.

The downside would be that this queue would be global (not yet sure if it is too bad).

It would work like this:

1. Initially we'll have it either null of undefined.
2. When user adds a folder, we assign this variable the result of calling main function that returns promise
3. Code in a function will run as usual using limited concurrency
4. If/When user adds more folders, we queue them up into existing promise, so they wait.

**The other way** might be to have a global (again) pool of files/folders to be processed and have a way to pick items from that pool and process them with a limited concurrency.

With this approach we don't batch anything and every items goes though the entire pipeline individually,
and it might not even end up in a database, because of filtering.
This might be slower because of the database pressure, but this will allow items to be added from multiple folders simultaneously. 

In both ways we might need to find a way to say when All folders are added, but not sure.

That's at least 2 ways I can think of ATM, there are might be more.

src/js/actions/LibraryActions.js
KeitIG: What do you mean ?

src/js/components/Settings/SettingsLibrary.react.js
KeitIG: W3C specs I guess

KeitIG:
I will play with [Queue](https://www.npmjs.com/package/queue) this afternoon and see how I can deal with those concurrency problems, then I will remove the WIP.

src/js/actions/LibraryActions.js
KeitIG: not working :p

KeitIG:
We should be good @YurySolovyov, I will check those build errors but the overall thing is doing what it's supposed to.

src/js/actions/LibraryActions.js
YurySolovyov: `async` does not give you anything if you're calling sync API.
`async` makes sense where you have some _really_ async function returning a promise or taking a callback.

src/js/actions/LibraryActions.js
YurySolovyov: can you post a gist of what you did?

KeitIG: I simple replaced 

```
        const paths = Promise.map(pathsToScan, (path) => {
            return statAsync(path).then((stat) => {
                return {
                    path,
                    stat
                };
            });
        });

        resolve(paths);
```

with your piece of code

YurySolovyov: ok, nevermind, maybe I'll take a look another time.

YurySolovyov:
Can't provide any better review ATM, will file an issue or PR if I find anything.

PR #332, opened 2017-03-28
Play automatically plays the first song when the player is stopped

### LINKED ISSUE #223: Play button behavior when playerStatus is "stop"
We do not allow the possibility to play a selected track using the play button. It's not user friendly IMO. 

To reproduce it select a track and just click the play button. 
- Expected behavior : Play the sound 
- Actual behavior : nothing happen. (and the play button change to pause even with no song playing, maybe another issue?)

### LINKED ISSUE #226: Current song stays in buffer when flushing queue
When you flush the queue and clic next, the current song stop, right
But when you click on play, the last current song resumes instead of playing the first track of the playlist/song list

KeitIG:
Didn't really think this would be that easy.

fix #223 
fix #226

### DIFF
--- a/src/js/actions/PlayerActions.js
+++ b/src/js/actions/PlayerActions.js
@@ -20,8 +20,11 @@ const audioErrors = {
 const playToggle = () => {
     const { paused } = Player.getAudio();
     // TODO (y.solovyov | KeitIG): calling getState is a hack.
-    const { queue } = store.getState();
-    if (paused && queue.length > 0) {
+    const { queue, playerStatus } = store.getState();
+
+    if(playerStatus === 'stop') {
+        start();
+    } else if (paused && queue.length > 0) {
         play();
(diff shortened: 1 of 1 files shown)


src/js/actions/PlayerActions.js
```diff
@@ -54,8 +57,13 @@ const start = (_id) => {
     // TODO (y.solovyov | KeitIG): calling getState is a hack.
     const { tracks, tracksCursor } = store.getState();
     const queue = [...tracks[tracksCursor].sub];
+
+    if(queue.length === 0) return;
+
+    const trackId = _id || queue[0]._id;
```
YurySolovyov: is there a case where both will be absent?

KeitIG: Theoretically, no, but if so, the line 61 should have caught this case.

PR #333, opened 2017-03-30
Fix tooltip position

KeitIG:


KeitIG:
Why is Travis still requested :x

PR #334, opened 2017-04-01
Fix cover fetch when it has single or doublequotes in its path

KeitIG:


YurySolovyov:
Aren't there node path's helpers that will do the job?

KeitIG:
didn't find any

KeitIG:
god, I don't know why Travis is still building!

KeitIG:
k, it should be fixed

YurySolovyov:
We may need to disable it on it's dashboard

PR #350, opened 2017-07-16
Fix shuffle rollback when shuffle is already enabled during a SELECT_AND_PLAY

### LINKED ISSUE #348: Re-implement shuffle system
The shuffle system is buggy when you unshuffle from other views. Try to think about a way to re-implement it in a simpler way.

KeitIG:
More a bugfix than a re-implementation.

Fix #348

### DIFF
--- a/src/js/reducers/player.js
+++ b/src/js/reducers/player.js
@@ -130,7 +130,7 @@ export default (state = {}, payload) => {
                 };
             }
 
-            // Unshuffle the queue by restoring the initial queu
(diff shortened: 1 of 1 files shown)


src/js/reducers/player.js
```diff
@@ -130,6 +130,7 @@ export default (state = {}, payload) => {
                 };
             }
 
+            // Unshuffle the queue by restoring the initial queu
```
YurySolovyov: `queue`

PR #355, opened 2017-08-28
Fix PageDown, PageUp, End and Start keys on TracksList

### LINKED ISSUE #306: Page up, page down, home, end in library view
We should add support for these keys in the library view, for quicker movement without a mouse. Page up/down would move one page up/down. Home would go to the top of the library. End would go to the bottom.

KeitIG:
Fix  #306

### DIFF
--- a/src/js/components/Shared/TracksList.react.js
+++ b/src/js/components/Shared/TracksList.react.js
@@ -49,7 +49,7 @@ export default class TracksList extends Component {
 
         // TODO (y.solovyov | KeitIG): TrackListHeader component?
         return (
-            <div className='tracks-list-container' tabIndex='0'>
+            <div className='tracks-list-container'> {/* there used to be a tabIndex={ 0 }, I don't remember why */}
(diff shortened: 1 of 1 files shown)


YurySolovyov:
I guess tabIndex was for focusing plain divs or something

KeitIG:
I believe it was for some custom behaviors, but I cannot remember what exactly. But it was blocking the default behavior of scrollable divs, some I am a bit annoyed.

KeitIG:
Yeah right, arrow keys do not work anymore. I will fix that.

KeitIG:
So yeah, I attached the whole event to  `document` instead of the container, it works, I hope I didn't miss a case, lgtm though.

YurySolovyov:
btw, there is a https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView for the case where selected row is outside the viewport and you want to make sure it is shown.

KeitIG:
Interesting, I will play with it to see what it can do.

Thx for the review!

Issue #349, opened 2017-07-13
Update react-router to v4


KeitIG:
I thought it would be a piece of cake. Hell I was wrong.

help: https://codeburst.io/react-router-v4-unofficial-migration-guide-5a370b8905a

PR #359, opened 2017-09-04
Misc

KeitIG:
- [x] Rename `src/js` to `src/ui`
- [x] Two indent spaces vs four? (the debate is open, I usually use 4, but most of projects I have seen use 2, in JS at least, and it allows more html nesting)
- [ ] ~Update `react-router` to v4 (#349)~
- [ ] ...


YurySolovyov:
2 spaces :+1:, given we have a lot of whitespace anyway

YurySolovyov:
Also, I'd take a look at some of the long lines wrapping we did, maybe some will fit into line length limit now.

YurySolovyov:
Also, IMO it would be better to update react router in a separate pr. Easier to review

KeitIG:
Well, I think we're good now. There were some component that could have been splitted, I will do it later when I have time. Codebase looks better anyway.

PR #361, opened 2017-09-05
Prevent mainWindow from opening new windows (fix #360)

### LINKED ISSUE #360: Middle Mouse Click UI Improvements
First of all great app using it a while know and it works fine so far :).
But i have one suggestion. Have you ever tried to click on a ui component with the middle mouse button?
On some components (i guess "<a>" tags or some kind of component with link interaction or redirect) it opens the new "screen/view" in another window instance (see screenshot)

image

Thats all from a single instance of the application. When you click on the playlist tab for example you get a screen/window like this.

image

Its now something to important, but i think it should be fixed because it causes unexpected behavior for the user.
I only know this because of a electron application i work on, so the same thing happend to me xD

KeitIG:


YurySolovyov:
how about https://electron.atom.io/docs/api/app/#appmakesingleinstancecallback ?

KeitIG:
We use it already https://github.com/KeitIG/museeks/blob/360-prevent-new-windows/src/main-process/main.js#L27

Apparently it's not enough, or I am using it wrong.

YurySolovyov:
FYI 1.7 has some problems with it ATM, not sure about 1.8 https://github.com/electron/electron/pull/10518

MarshallOfSound:
@YurySolovyov 1.8 has the same issues 👍 that PR fixes both

YurySolovyov:
FWIW [v1.7.8](https://github.com/electron/electron/releases/tag/v1.7.8) is released

PR #374, opened 2017-12-31
Fix production build

### LINKED ISSUE #373: Fix production build
Production build is ok, you can see it after compiling the app, `redux-logger` is still active and the bundle is huge.

KeitIG:
Fix #373


package.json
```diff
@@ -122,13 +122,13 @@
...
-    "minimist": "1.2.0",
     "node-sass": "4.7.2",
     "redux-logger": "^3.0.6",
     "sass-loader": "6.0.6",
     "style-loader": "0.19.0",
     "svg-inline-loader": "0.8.0",
     "url-loader": "0.6.2",
-    "webpack": "3.8.1"
+    "webpack": "3.8.1",
+    "webpack-merge": "^4.1.1"
```
YurySolovyov: let's use exact versions in `package.json`

webpack.dev.js
```diff
@@ -0,0 +1,18 @@
...
+const common = require('./webpack.common.js');
+
+
+module.exports = merge(common, {
+  devtool: 'inline-source-map',
+  plugins: [
+    new CleanTerminalPlugin(),
+    new webpack.DefinePlugin({
+      'process.env': {
+        NODE_ENV: JSON.stringify('development'),
```
YurySolovyov: why do we need JSON.stringify here? I think `"'development'"` or `\""development"\"` might work too

YurySolovyov:
quite ironic that https://github.com/KeitIG/museeks/commit/608fc318d56189c3eb43cce02451a29b4e15d935 was "Fix production build" too

KeitIG:
Yeah, this commit message is somewhat wrong haha. The problem when you want to update packages but do not really read all their changelogs.

package.json
KeitIG: A question about this: is that still necessary to use exact versions when we have a lock file?

YurySolovyov:
I usually say something like "Bump deps" or "Update React" etc.

package.json
YurySolovyov: Hmm, fair point, but I'd still prefer exact ones. Up to you.

KeitIG:
I made the needed changes. About Yarn vs NPM, do you have a preference? I switched to Yarn a few weeks ago, but I do not still know if it was a right move.

YurySolovyov:
I prefer npm over yarn, less stuff to involve.

package.json
```diff
@@ -106,14 +106,14 @@
...
     "babel-plugin-transform-class-properties": "6.24.1",
     "babel-plugin-transform-object-rest-spread": "6.26.0",
     "babel-preset-env": "1.6.1",
     "babel-preset-react": "6.24.1",
-    "babili-webpack-plugin": "0.1.2",
-    "clean-terminal-webpack-plugin": "^1.0.1",
+    "clean-terminal-webpack-plugin": "1.0.1",
     "css-loader": "0.28.7",
-    "electron": "^1.7.9",
+    "electron": "1.7.9",
```
YurySolovyov: lastest is `1.7.10` by now

Issue #378, opened 2018-01-18
Reorganize main-process


KeitIG:
A small collection of problems I found when working on the MacOS integration

- [x] Move IPC event name to a enum file
- [x] Retrink how we call integrations as it is really procedural
- [x] Double check all app and webContents events as they are dispatched through multiple files and are not predictable

PR #379, opened 2018-01-18
macOS integration

### LINKED ISSUE #337: macOS integration issues
I had some time to test Museeks on Mac, and it's not perfect

- [x] Museeks interrupts logout (cf. screenshot)
- [x] Museeks does not show if minimized to tray + click it's in the dock
- [x] Right-click on the dock icon + quit doesn't quit the app
- [x] Native window controls
- [x] Fix-multiselection #368 
- [x] Native font

screen shot 2017-04-04 at 10 33 20

### LINKED ISSUE #368: ctrl-click triggers context-menu on macOS
Maybe we should use the cmd event modifier instead of ctrl on macOS for multi-selection

KeitIG:
Do not mind the mess in main-process, #378 should solve it.

Fix #337 
Fix #368


src/ui/components/Header/Header.react.js
```diff
@@ -102,7 +96,7 @@ const mapStateToProps = ({ player }) => ({
   shuffle: player.shuffle,
   queue: player.queue,
   queueCursor: player.queueCursor,
-  useNativeFrame: config.get('useNativeFrame'),
+  showTopHeader: os.platform() !== 'darwin' && !config.get('useNativeFrame'),
```
YurySolovyov: can't we group these together to get rid of negations? Like:
```js
showTopHeader: !(os.platform() === 'darwin' && config.get('useNativeFrame'))
```

src/ui/styles/_header.scss
```diff
@@ -56,18 +56,22 @@ header {
...
+    width: 220px;
     display: flex;
     align-items: center;
+    padding-right: 10px;
   }
 
   /* .windows-controls */
   .col-player-infos {
-    width: calc(100% - 400px);
+    width: calc(100% - 440px);
```
YurySolovyov: Slightly off-topic, but I think we might eventually want to use [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/var)

KeitIG: Totally, I planned to refactor these SCSS files with variables and other best pratices

src/ui/components/Header/Header.react.js
KeitIG: I think the correct condition would be `!(os.platform() === 'darwin' || config.get('useNativeFrame'))`, which I find quite less understandable than the current one

KeitIG: I will add a comment though

src/ui/styles/_header.scss
YurySolovyov: Ideally I'd prefer CSS variables since they are first class in the browser where SCSS vars only exist in the code

src/ui/components/Header/Header.react.js
vprigent: Can we extract all those platform related conditions somewhere specific for the whole application ?
This would remove all those messy conditions `!os.specific && foobar`

KeitIG: Yeah we might want that at some point indeed

PR #388, opened 2018-02-06
Dark tray icon option

a2nt:
Updated commit to follow original code styling.
screenshot from 2018-02-06 18-32-44
screenshot from 2018-02-06 18-33-00
screenshot from 2018-02-06 18-33-19


src/main-process/main.js
```diff
@@ -66,7 +66,7 @@ app.on('ready', () => {
     '48': nativeImage.createFromPath(path.join(logosPath, 'museeks-48.png')),
     '32': nativeImage.createFromPath(path.join(logosPath, 'museeks-32.png')),
     'ico': nativeImage.createFromPath(path.join(logosPath, 'museeks.ico')),
-    'tray': nativeImage.createFromPath(trayIconPath).resize({ width: 24, height: 24 }),
+    'tray': nativeImage.createFromPath(trayIconPath),
```
KeitIG: So I remember adding this line when I was on Elementary because otherwise the. icon would actually take 256px in the top bar.

I will try to reproduce it.

a2nt: looks like Skype (the latest version is based on nodejs) has the same issue, also MailSpring, Harmony just need to add some borders cuz it's icon is too big, but Slack works great.

package.json
```diff
@@ -80,10 +80,12 @@
     "globby": "7.1.1",
     "keymirror": "0.1.1",
     "level-js": "2.2.4",
+    "leveldown": "^3.0.0",
```
YurySolovyov: please, use exact versions of deps

src/main-process/main.js
```diff
@@ -15,7 +15,7 @@ const nativeImage   = electron.nativeImage;
 const BrowserWindow = electron.BrowserWindow; // Module to create native browser window.
 
 const appRoot = path.resolve(__dirname, '../..'); // app/ directory
-const srcPath = path.join(appRoot, 'src'); // app/src/ directory
+const srcPath = path.join(appRoot, 'src'); // app/src/ directoryz
```
YurySolovyov: typo?

a2nt: yup, sorry a sec

a2nt:
@YurySolovyov requested fixtures are done

YurySolovyov:
Thanks, but I meant exact versions for all (changed/updated/added) deps.

a2nt:
@YurySolovyov done

src/main-process/main.js
```diff
@@ -129,6 +136,26 @@ app.on('ready', () => {
   // integrations
   const integrations = new IntegrationManager(mainWindow);
   integrations.enable();
+
+  // Update tray icon for gnome-shell
+  if(process.platform === 'linux') {
```
YurySolovyov: we have a subdir for platform-specific integrations, currently there is only win32 file, but nothing is preventing you to add linux.js and keep this file more generic

src/main-process/main.js
```diff
@@ -129,6 +136,26 @@ app.on('ready', () => {
...
   const integrations = new IntegrationManager(mainWindow);
   integrations.enable();
+
+  // Update tray icon for gnome-shell
+  if(process.platform === 'linux') {
+    const ps = require('ps-node');
+    // detect gnome-shell process
+    ps.lookup({
+      command: 'gnome-shell',
+    }, (err, resultList ) => {
```
YurySolovyov: shouldn't linter catch this weird spacing ?

a2nt: nope, checkout npm run lint:js don't see any issues with this spacings

src/main-process/main.js
a2nt: fixed

package.json
```diff
@@ -99,7 +101,10 @@
     "react-svg-inline": "2.1.0",
     "redux": "3.7.2",
     "redux-thunk": "2.2.0",
+    "sass-lint-fix": "1.12.1",
     "semver": "5.5.0",
+    "strip-ansi": "4.0.0",
+    "table-parser": "0.1.3",
```
YurySolovyov: Why to you have to add `strip-ansi` and `table-parser` directly when they are deps of `ps-node` ?

a2nt: sass-lint-fix depends on these packages.
You can use: sass-lint --fix to autofix spacing issues as long as eslint --fix

YurySolovyov: I still don't get it. if `sass-lint-fix` depends on them, npm will install them anyway.

YurySolovyov:
Posting here so that it does not get lost:

I still don't get why do you need to add `sass-lint-fix` deps explicitly, if `sass-lint-fix` depends on them, npm will install them transitively anyway.

package.json
a2nt: before u have asked to include all requirements

KeitIG:
I am thinking about removing the Tray in a future version of Museeks. Moreover, if the issue comes from the color of the tray icon, we should solve it in a nice way (having a better Tray Icon), like using the monochrome one we use on MacOS.

Sorry for the time it took, and thank you for the contribution 🙌

martpie:
Oh well, it's here.

package.json
YurySolovyov: If you mean "please, use exact versions of deps" from the comment above, I meant that we should use exact versions of deps in a semver sense - i.e. not using any kind of version specifiers that allow "automatic" version bumps, e.g. instead of `~1.0.3` or `^3.14.1` we should just use `1.0.3` and `3.14.1`

PR #394, opened 2018-02-06
sass-lint

KeitIG:
Fix #394


src/ui/styles/main.scss
```diff
@@ -1,72 +1,56 @@
...
 
-/*
-|--------------------------------------------------------------------------
-| Views
-|--------------------------------------------------------------------------
-*/
+@import "view-common";
+@import "view-library";
+@import "view-playlists";
+@import "view-settings";
```
YurySolovyov: Don't we use single quotes in JS?

YurySolovyov:
Have you considered using `stylus` or `postcss` ? I hate sass for native deps

src/ui/styles/main.scss
KeitIG: it's a scss file here :p

KeitIG:
@YurySolovyov Yeah, you are right, I like stylus a lot, and someone told me postcss was amazing. I should give them a try.

Issue #175, opened 2016-08-26
Reorderable playlists


martpie:
As _simple_ as that.

~~Blocked by https://github.com/KeitIG/museeks/issues/112~~

KeitIG:
Not really blocked by #112 actually.

martpie:
FINALLY #470

PR #411, opened 2018-03-06
Sorting

KeitIG:
🎉 I gotta say that was a rather funny one to develop.

screen shot 2018-03-06 at 12 08 53
screen shot 2018-03-06 at 12 08 59
screen shot 2018-03-06 at 12 10 04


src/ui/components/Shared/TracksList.react.js
```diff
@@ -329,20 +330,12 @@ export default class TracksList extends Component {
   }
 
   render() {
-    const tracks = [...this.props.tracks];
+    const { tracks, type } = this.props;
 
-    // TODO (y.solovyov | KeitIG): TrackListHeader component?
     return (
-      <div className='tracks-list-container'> {/* there used to be a tabIndex={ 0 }, I don't remember why */}
+      <div className='tracks-list'> {/* there used to be a tabIndex={ 0 }, I don't remember why */}
```
YurySolovyov: > there used to be a tabIndex={ 0 }, I don't remember why

To be able to Tab over the fields?

src/ui/components/Shared/TracksList.react.js
```diff
@@ -329,20 +330,12 @@ export default class TracksList extends Component {
...
         <KeyBinding onKey={this.onKey} preventInputConflict />
-        <div className='tracks-list-header'>
-          <div className='track-cell-header cell-track-playing' />
-          <div className='track-cell-header cell-track'>Track</div>
-          <div className='track-cell-header cell-duration'>Duration</div>
-          <div className='track-cell-header cell-artist'>Artist</div>
-          <div className='track-cell-header cell-album'>Album</div>
-          <div className='track-cell-header cell-genre'>Genre</div>
-        </div>
+        <TracksListHeader enableSort={type === 'library'} />
```
YurySolovyov: Sorting is not supported in playlists?

src/ui/components/Shared/TracksListHeaderCell.react.js
```diff
@@ -0,0 +1,68 @@
...
+    return (
+      <React.Fragment>
+        <div className='col-name'>
+          {props.title}
+        </div>
+        {props.icon &&
+          <div className='col-icon'>
+            <Icon name={props.icon} />
+          </div>
+        }
```
YurySolovyov: why not just use plain `if`?

src/ui/components/Shared/TracksListHeaderCell.react.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+
+  render() {
+    const { sortBy, className } = this.props;
+    const classes = classnames('track-cell-header', className, {
+      sort: sortBy,
+    });
+
+    if (sortBy) {
```
YurySolovyov: do you think it might be better / more clear to always render the button, but have it in "unsorted" state by default instead of rendering it conditionally?

package.json
```diff
@@ -85,6 +85,7 @@
     "keymirror": "0.1.1",
     "level-js": "2.2.4",
     "linvodb3": "3.25.1",
+    "lodash.orderby": "^4.6.0",
```
YurySolovyov: exact version

src/ui/components/Shared/TracksList.react.js
KeitIG: So yeah, that is one of the problem I faced. The thing with playlist is do not want them to be sortable, I want them to be re-orderable (#175).

Let's imagine we have sortable and re-orderable playlist, switching from one mode to the other will introduce a lot of complexity. I checked other players, and most of them do not allow you to sort playlists.

src/ui/components/Shared/TracksListHeaderCell.react.js
KeitIG: How that?

src/ui/components/Shared/TracksListHeaderCell.react.js
KeitIG: I went with the semantic approach here, a disabled button does not really make a lot of sense I think.

src/ui/components/Shared/TracksList.react.js
KeitIG: I think this was related to the keyboard event, but I cannot reproduce any wrong behavior.

src/ui/components/Shared/TracksListHeaderCell.react.js
YurySolovyov: I thought you could use plain if there, nevermind

src/ui/components/Shared/TracksList.react.js
YurySolovyov: Fair.

KeitIG:
🚀

Issue #369, opened 2017-12-03
Performance on huge libraries


SirWindfield:
I just wanted to ask how the performance is when libraries are used that contain 10000+ songs. I added my folders and museeks is still scanning for about 7 hours now...

I checked the console but there weren't any errors displayed.

I took a quick look into the state and noticed that you store every song there. Isnt that quite RAM eating? Not sure if it is enough information to increase the ram usage up by a lot.

KeitIG:
Hello :)

I do have a library of 6000 mp3 (around 20GB), and it is being scanned in less than a minute, which is quite fair :smile:

It is strange that you get 7 hours of scan without any warning. Are you using a HDD or a SSD? What is the exact size (in weight) of your library?

SirWindfield:
Thanks for the reply!

My library is about 200GB large. Mostly a mix of flac,wav and mp3 files. The number of songs is around 30000. 
I made sure to start museeks directly in dev mode and now I get an error. Though that the stack trace isn't that useful ([gist](https://gist.github.com/SirWindfield/c38521231371000107482aef531027dc)).

YurySolovyov:
I think problem is somewhat in the way we get metadata - some code paths may cause reading a significant portion of the file before we get the metadata.

@SirWindfield do you have a large number of mixes like when entire album/mix is in one file?

SirWindfield:
@YurySolovyov No I do not have. The largest files are usually podcasts. 
Can you help me out real quick? I downloaded the repo, installed electron globally and want to run it using `yarn run dev` but the electron instance doesn't load from the provided dev-server. Did I forget something?

KeitIG:
The only things I see here is some files are not having metadata headers but this is fine and fallbacks are ready for this scenario.

It would be interesting that I build the latest version of Museeks from sources and make you test it to see if you can still reproduce it.

(I just need to finish what I am working on and I hope it will be finished in the next days)

SirWindfield:
Thanks. I will gladly test it out. Best would be to push the branch upstream and let me build itself because I am currently running on a metered connection (isp is changing some cables down the road).

What about the error caused by your `musicmetadata` dependency? Looks like it tries to format some nonexistent tags for some odd reason.
If I remember correctly, `yarn` warned me that I should use `music-metadata` instead. Maybe the author pushed some updates onto that library instance concerning the bug. Might be worth an investigation.

KeitIG:
[...]

PR #434, opened 2018-04-11
TracksList performance improvements

KeitIG:
That's a first step. One of the main benefit is we don't see anymore some blank space when scrolling really fast.

#369 

Before 

screen shot 2018-04-11 at 15 24 22

After 

screen shot 2018-04-11 at 16 27 53

### DIFF
--- a/src/ui/components/Shared/TracksList.react.js
+++ b/src/ui/components/Shared/TracksList.react.js
@@ -4,7 +4,6 @@ import PropTypes from 'prop-types';
 import KeyBinding from 'react-keybinding-component';
 
 import TrackRow from './TrackRow.react';
-import PlayingIndicator from './PlayingIndicator.react';
 import CustomScrollbar from './CustomScrollbar.react';
 import TracksListHeader from './TracksListHeader.react';
 
@@ -13,7 +12,6 @@ import * as PlaylistsActions from '../../actions/PlaylistsActions';
 import * as PlayerActions from '../../actions/PlayerActions';
 import * as QueueActions from '../../actions/QueueActions';
 
-import Player from '../../lib/player';
 import * as utils from '../../utils/utils';
 import { isCtrlKey } from '../../utils/utils-platform';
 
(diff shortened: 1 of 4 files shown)


src/ui/components/Shared/TracksList.react.js
```diff
@@ -40,25 +43,27 @@ export default class TracksList extends Component {
...
   componentDidMount() {
     this.renderView = document.querySelector('.tracks-list-render-view');
   }
 
-  scrollTracksList() {
-    this.setState({ scrollTop : this.renderView.scrollTop });
+  onScroll() {
+    const tilesScrolled = Math.floor(this.renderView.scrollTop / TILE_HEIGHT);
+
+    if (this.state.tilesScrolled !== tilesScrolled) {
```
YurySolovyov: Are you sure this is not checked by React itself?
Like, does React re-render components if you set state with exactly the same values?

KeitIG: Unfortunately yes, I just used this to avoid having to implement a complex `shouldComponentUpdate` https://stackoverflow.com/questions/24718709/reactjs-does-render-get-called-any-time-setstate-is-called#24719289

YurySolovyov: Weird, [React docs say that setState is async inside event handlers](https://reactjs.org/docs/faq-state.html#when-is-setstate-asynchronous). 
`onScroll` seems like event handler to me

KeitIG: `setState` is async indeed, but if you `this.setState({a: 5})` and `this.state.a` already equals 5, `render()` will be re-triggered.

YurySolovyov:
BTW, for scrolling, have we considered using [passive event listeners](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners)?
I think they are more performant for cases where you don't need to `e.preventDefault()`

KeitIG:
I did consider passive events, but at the time I checked, It actually lost perfs (do not ask why). 

And we are blocked by https://github.com/facebook/react/issues/6436 to implement in cleanly.

PR #455, opened 2018-08-19
Typescript

KeitIG:


KeitIG:
Next steps will be to have the main-process code in TypeScript too, but not urgent.

When https://github.com/palantir/tslint/pull/3641 will be merged into TSLint, we will be able to use it for JS files, that should be able to mitigate eslint's drop.

KeitIG:
There is probably a couple of bugs left, I did not test the whole application, but overall it's great. Actions still need to be typed.

PR #472, opened 2019-02-28
m3u support

martpie:
- [x] m3u import
- [x] Prevent a `.m3u` to be imported twice
- [x] m3u export

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4223,6 +4223,21 @@
         "postcss": "^7.0.14"
       }
     },
+    "id3v2": {
+      "version": "0.2.2",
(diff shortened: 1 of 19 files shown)


YurySolovyov:
Not sure what's the best timing, but it might be useful to start phasing out bluebird in favor of native promises, mainly for performance (both runtime and launch) reasons.

There is a [Node API for promisification](https://nodejs.org/api/util.html#util_util_promisify_original) and standalone modules that implement some of the sugar bluebird provides. (see [p-all](https://github.com/sindresorhus/p-all) and [related](https://github.com/sindresorhus/p-all#related))

martpie:
yep, I planned to do that, it should not be too hard I think

PR #473, opened 2019-03-03
Added MPRIS module to handle native integration with dbus service on GNU/Linux

igorer88:
- Added dependencies packages in order to work.
- Added mpris-service, node-loader to load .node bin files and electron-rebuild.
- Added electron-rebuild script as museeks:rebuild.
- On mpris type definition I tried to implement it, but I had my doubts about it and I end up commenting on the lines to keep working. 
- Added "playback:stop" event to allow stop the current playlist.
- Changed utils file location to have a better code consistency. As recommended by @martpie 
- TODO: fetch cover, handle the seek bar, handle the volume.


martpie:
Well, the CI already gives us a nice indication of what needs to be changed. I hate native modules.

src/main/modules/mpris.ts
```diff
@@ -0,0 +1,107 @@
...
+
+class MprisModule extends ModuleWindow {
+
+  player = mpris({
+      name: 'museeks',
+      identity: 'Musseks Music Player',
+      canRaise: true,
+      supportedUriSchemes: ['file'],
+      supportedMimeTypes: ['audio/mpeg', 'application/ogg'],
+      supportedInterfaces: ['player', 'trackList']
```
martpie: I don't think we support `trackList` do we?

src/main/modules/mpris.ts
```diff
@@ -0,0 +1,107 @@
...
+  });
+
+  constructor (window: Electron.BrowserWindow) {
+    super(window);
+    this.platforms = ['linux'];
+  }
+
+	async load() {
```
martpie: you have an indentation problem. You should catch it by running `npm run lint:ts`

src/shared/utils.ts
```diff
@@ -5,7 +5,7 @@ import * as globby from 'globby';
 import * as mmd from 'music-metadata';
 import pickBy from 'lodash-es/pickBy';
 
-import { Track } from '../../shared/types/interfaces';
+import { Track } from './types/interfaces';
```
martpie: I think there was a misunderstanding. This file should stay where it was. Only extract the `fetchCover` (and the `parseBase64` I guess) to `src/shared/utils.ts`.

(only move those two function, not the entire file)

src/ui/actions/AppActions.ts
```diff
@@ -109,6 +109,10 @@ const init = async () => {
     await PlayerActions.next();
   });
 
+  ipcRenderer.on('playback:stop', async () => {
```
martpie: It doesn't need to be `async` if it does not return a `Promise` or if it does not use the `await` keyword.

package.json
```diff
@@ -77,10 +79,12 @@
     "css-loader": "2.1.0",
     "electron": "4.0.5",
     "electron-builder": "20.38.5",
+    "electron-rebuild": "^1.8.4",
```
martpie: same

package.json
```diff
@@ -36,6 +37,7 @@
     "level-js": "4.0.0",
     "linvodb3": "3.26.0",
     "lodash-es": "4.17.11",
+    "mpris-service": "^1.1.4",
```
martpie: Please use the exact version

package.json
```diff
@@ -77,10 +79,12 @@
     "css-loader": "2.1.0",
     "electron": "4.0.5",
     "electron-builder": "20.38.5",
+    "electron-rebuild": "^1.8.4",
     "file-loader": "3.0.1",
     "html-webpack-plugin": "3.2.0",
     "install": "0.12.2",
     "mini-css-extract-plugin": "0.5.0",
+    "node-loader": "^0.6.0",
```
martpie: same

martpie:
There are still conflicts to solve as well. I will investigate for the build issue and the mpris cover.

src/main/modules/mpris.ts
```diff
@@ -0,0 +1,107 @@
...
+import * as mpris  from 'mpris-service';
+const { app, ipcMain } = electron;
+
+class MprisModule extends ModuleWindow {
+
+  player = mpris({
+      name: 'museeks',
+      identity: 'Musseks Music Player',
+      canRaise: true,
+      supportedUriSchemes: ['file'],
```
martpie: You should try to add `data` here, it may make the dataUri cover work.

https://specifications.freedesktop.org/mpris-spec/2.0/Media_Player.html#Property:SupportedUriSchemes

src/main/modules/mpris.ts
```diff
@@ -0,0 +1,107 @@
...
+      supportedMimeTypes: ['audio/mpeg', 'application/ogg'],
```
martpie: Here I think you should import the `SUPPORTED_TRACKS_EXTENSIONS` variables from `ui/lib/app.ts`.

Then you can simply map it with a module like `mime-type` https://www.npmjs.com/package/mime-types

So it looks like:

```js
import * as mime from 'mime-types';

...
supportedMimeTypes: SUPPORTED_TRACKS_EXTENSIONS.map(mime.lookup);
```

src/main/modules/mpris.ts
igorer88: That got me thinking, as it doesn't play as long we don't push the play button on the player.
Maybe it needs to add the tracklist for the mpris module to play what's loaded on the player. :confused:

src/main/modules/mpris.ts
igorer88: The IDE didn't show me anything :confused:

src/shared/utils.ts
igorer88: Oh shit. Let's fix it.

package.json
igorer88: ok :+1:

src/main/modules/mpris.ts
igorer88: Ok. I'll try it. :+1:

src/main/modules/mpris.ts
igorer88: ok, I'll add this later. Let's focus on the fetch cover feature.

src/main/modules/mpris.ts
igorer88: I didn't have installed the ts linter. Sorry. It's solved xD

martpie:
@igorer88 Do not forget to push your changes if you made any ;)

martpie:
Good news, `mpris-service` is now pure-JS (but not released yet). I will simplify native deps requirement a lot.

martpie:
Currently working on it

martpie:
Let's close this in favor of #475

Issue #482, opened 2019-04-27
Empty imported m3u playlists


aDogCalledSpot:
I have some playlists I would like to import. Some lines seem to cause the import to fail. When it fails, the playlist is created but no songs are added (not even the ones that don't cause issues), I don't get any output in the terminal when running from it. I cannot tell what is causing this behavior since similarly complex file paths can either cause the issue or not.

Creating the playlist test1.m3u with only the one file path in it works:
`/home/spot/Media/Music/MusicBee/MusicBee Media/Tristam & Braken/Flight/Flight (Original Mix).mp3`

Creating the playlist test2.m3u with only this filepath does not work:
`/home/spot/Media/Music/MusicBee/MusicBee Media/Tristam & Braken/Monstercat 012 - Aftermath/01 Flight.mp3`

Both files exist and can be recognized correctly by other programs.

Adding/removing new lines at the end of the file does not seem to change anything.

PR #485, opened 2019-04-28
Fix failing m3u imports

aDogCalledSpot:
Fixes https://github.com/martpie/museeks/issues/482

Carriage returns at the end of the lines were messing things up. This checks if a carriage return is at the end of the current line and removes it if it finds one.

### DIFF
--- a/src/ui/utils/utils-m3u.ts
+++ b/src/ui/utils/utils-m3u.ts
@@ -18,7 +18,7 @@ export const parse = (filePath: string): string[] => {
     const decodedContent = iconv.decode(content, encoding);
 
     const files = decodedContent
-      .split('\n')
+      .split(/\r?\n/)
       .reduce((acc, line) => {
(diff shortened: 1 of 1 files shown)


src/ui/utils/utils-m3u.ts
```diff
@@ -24,6 +24,11 @@ export const parse = (filePath: string): string[] => {
           return acc;
         }
 
+        // Remove trailing carriage returns
+        if (line.charAt(line.length - 1) === '\r') {
+          line = line.substring(0, line.length - 1);
```
YurySolovyov: I think it is better to just split by regex on line 21:
```js
.split(/\r?\n/)
```

aDogCalledSpot: Great idea!

martpie:
Hello there, thank you for the PR, I will try to give it a look asap :)

Issue #495, opened 2019-07-08
Suggestion: Double clicking playlist starts playing


MrDoctorKovacic:
Great work, this app is beautiful and handles my 10K+ song library with no issues. Has become my goto Linux music player!

Small QOL change, could we make it so double clicking a playlist in the sidebar will start playing it? This feature is present in iTunes on MacOS.

I'll take a swing at adding it myself, but wanted to put it on the radar. Thanks!

martpie:
Makes sense and should not be too hard to implement, I'll have a look!

PR #496, opened 2019-07-08
Adding Double Click to play (#495)

MrDoctorKovacic:
#495

I'm pretty new to both React and the whole git system, please let me know if anything here is less-than-ideal

### DIFF
--- a/src/ui/components/PlaylistsNavLink/PlaylistsNavLink.tsx
+++ b/src/ui/components/PlaylistsNavLink/PlaylistsNavLink.tsx
@@ -1,11 +1,9 @@
 import * as React from 'react';
 import { NavLink } from 'react-router-dom';
 
-import * as app from '../../lib/app';
-import * as PlayerActions from '../../actions/PlayerActions';
-import * as styles from './PlaylistsNavLink.css';
+import * as PlaylistActions from '../../actions/PlaylistsActions';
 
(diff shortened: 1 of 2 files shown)


src/ui/components/PlaylistsNavLink/PlaylistsNavLink.tsx
```diff
@@ -1,15 +1,30 @@
...
   playlistId: string;
   onContextMenu: (playlistId: string) => void;
 }
 
 export default class PlaylistsNavLink extends React.Component<Props> {
+  // Start playing playlist on double click (Issue #495)
+  handleDoubleClick = async () => {
+    try {
+      const playlist: PlaylistModel = await app.models.Playlist.findOneAsync({ _id: this.props.playlistId });
+      const tracks: TrackModel[] = await app.models.Track.findAsync({ _id: { $in: playlist.tracks } });
```
martpie: This (all async operations like this one) should be part of an action, how about `PlaylistActions.play()` or `PlaylistActions.startPlayback()`?

https://github.com/martpie/museeks/blob/master/src/ui/actions/PlaylistsActions.ts

MrDoctorKovacic: Okay, great idea!

MrDoctorKovacic:
No worries at all, thank you for looking through this. Something along these lines?

Issue #494, opened 2019-07-04
New playlists dont show tracks until renamed


jdbevan:
macOS Mojave v10.14.5

Museeks 0.11.0

I can reproducably select some tracks, right click, create new playlist, then navigate to the playlist and no tracks are displayed. Once I've right clicked on the playlist name and renamed it, then the tracks appear. The recording below contains a word in the search box in the top-right which I thought might be the source of the bug, but I see the same behaviour with an empty search box

![museeks-new-playlist](https://user-images.githubusercontent.com/1391574/60670510-c4612900-9e68-11e9-8b66-2e9dfa116015.gif)

martpie:
Oh oh :see_no_evil: let me try to reproduce it

PR #500, opened 2019-07-13
Fixing empty playlist on context menu create

MrDoctorKovacic:
#494 

Enjoy your weekend!


martpie:
You create a playlist, but not adding any files in it anymore, which breaks the feature 😄 

The problem is probably coming from somewhere else, I'll investigate

MrDoctorKovacic:
Thanks for looking, but are you sure? This worked when I tested it.

Another option is to manually refresh after the original 2 lines

martpie:
Ah wait, I may have missed sth

martpie:
you are actually absolutely right

PR #502, opened 2019-07-26
Update README.md

scriptspry:
Added helpful note that museeks is available for installing as a cask on homebrew.


martpie:
Mmmh, I need to think about it. While it is true, I am not the maintainer of the cask itself (I updated it a couple of times, but I don't do it automatically at each release) 🤔

scriptspry:
Makes sense. Shall I close this pull request and create an issue requesting the cask instead?

martpie:
I think it makes sense to request an offical cask, maybe we can automate this process on releases👍

PR #505, opened 2019-08-18
Give GitHub Actions a try

martpie:


.github/workflows/qa.yml
```diff
@@ -0,0 +1,38 @@
+name: Node CI
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
YurySolovyov: I'd better use fixed version/release if possible. (same idea as for JS packages)

PR #507, opened 2019-08-27
Update to Electron 6

martpie:
Promises all the way.


src/ui/actions/LibraryActions.ts
```diff
@@ -244,49 +245,45 @@ export const add = async (pathsToScan: string[]) => {
...
-    const result = dialog.showMessageBox(app.browserWindows.main, {
+    const result = await dialog.showMessageBox(app.browserWindows.main, {
       buttons: ['Cancel', 'Reset'],
       title: 'Reset library?',
       message: 'Are you sure you want to reset your library ? All your tracks and playlists will be cleared.',
       type: 'warning'
     });
 
-    if (result === 1) {
+    if (result.response === 1) {
```
YurySolovyov: (not directly related, but)
I'd move this `1` to some kind of enum/object/constant to read:
```js
if (result.response === DialogResults.SUCCESS) { ... }
```
or something like that

martpie: I would expect Electron to provide these enums. I will check.

PR #508, opened 2019-10-03
Fixes quick search for album or artist in right click menu

vprigent:
Use headerStyles for search input in querySelector


martpie:
Oh oh 🙈

Issue #465, opened 2018-12-13
touchscreen controls


DennisDray:
the volume control button needs to be tapped to show volume slider
which mutes sound
so need to press again to unmute
moreover, once the slider appears, the control buttons stay gone

best is to put the slider in place of the mute button

martpie:
Never thought about touchscreen, thank your for the feedback.

I’ll try to think about a solution for that.

dreaddr:
Yeah, this would be nice, a player that works on both mouse and touchscreen mode.

PR #509, opened 2019-10-08
Change volume control position

### LINKED ISSUE #351: Volume bar opens to left of volume icon and covers the play buttons
**The volume bar opens at the left of the volume icon and covers the play buttons** until you move the mouse away. Sometimes you want to click pause you move your mouse over the volume icon and you need to make a second move to close the volume bar that pops up. Being on the right side of volume icon makes it more simple.

screenshot from 2017-07-23 13-20-02
screenshot from 2017-07-23 13-21-43

My suggestion is to make it open at the right of the icon 😉 

**Specs**: 
Museeks 0.8.1 (Installed via AUR) -- OS: _Arch Linux_ -- Kernel: _x86_64 Linux 4.11.7-1-ARCH_

markatk:
This changes the volume control position to the right side of the playback controls and makes it permanently visible. It also fixes the initial volume slider position.

image

I think it is better that way because you don't need an additional click for an pop-over and I did quite often want to advance a song but accidental hovered the volume icon and couldn't access the buttons underneath.

It fixes #351 and #465


martpie:
Thanks a lot for the PR, I'll test it tonight. 🙌

The main reason I did not do it before was when the window is small, it gets quite tricky to fit everything in one line.

martpie:
we have a little issue I'm afraid :) maybe positionning it below like a tooltip with a box-shadow could be a better solution

Screen Shot 2019-10-08 at 21 07 23

markatk:
Ah I see, I just tested it on linux and windows, where the window buttons where not in the playback area. I'll try to find a solution for this.

markatk:
I have changed the volume slider to open a popover, when using macOS and there is not enough space. This way it is not the same all the time, but I still think its best to just show the slider in the title bar when there is enough space.

When the window is too small to show the slider in the title bar:
image

martpie:
Looks better, there's still one little issue: when dragging the volume left or right, the window sometimes moves. You may need to tweak `-webkit-app-region` somewhere :)

markatk:
I added the css flag but don't know if this is enough for all cases.

martpie:
Hey, thanks a lot for this again, the code has changed a lot since this PR, so I think it's better to close it. It's definitely on my radar and plan to improve this by the next release (soon)

Issue #511, opened 2020-02-01
Playback Speed


ShipRekt101:
I can't set any playback speed between 1 and 2. There is no change until 2, which then suddenly speeds up the song.

PR #512, opened 2020-02-16
use parseFloat instead of parseInt

qcasey:
Solves #511


martpie:
That looks quite correct :) I'll try to publish a release soon with your fix, thanks!

Issue #504, opened 2019-07-31
Bug : When create playlist and add song to playlist have to restart app


modernduck:
(Version 0.11.1 (0.11.1))

Hi When create playlist and add song to playlist; It didn't show up on playlist (but you could play whole playlist) But when I restart the app I could see list of tracks in the playlist again.

martpie:
Thank you for the report, I'll investigate!

modernduck:
> Thank you for the report, I'll investigate!

I think it happen  like this too
when you remove track in playlist to be zero  => then add new track => won't see track in playlist

martpie:
Repro:

- Create new playlist
- Go to tracks list
- Add tracks to the playlist
- Go back to the playlist screen
- Playlist is not refreshed

PR #513, opened 2020-02-16
refresh after adding tracks to Playlist

qcasey:
This is fixed when calling 
`    await refresh();`
in the addTracks function, exactly like when calling rename() or remove()


qcasey:
this is for #504 by the way

src/ui/actions/PlaylistsActions.ts
```diff
@@ -123,6 +123,7 @@ export const addTracks = async (_id: string, tracksIds: string[], isShown?: bool
     const playlist = await app.models.Playlist.findOneAsync({ _id });
     const playlistTracks = playlist.tracks.concat(tracksIds);
     await app.models.Playlist.updateAsync({ _id }, { $set: { tracks: playlistTracks } });
+    await refresh();
```
martpie: I noticed this bug as well but I am not sure this is the "right" way to fix it, I'll need to investigate a tiny bit before accepting this.

martpie:
By the way, I've fixed the checks on master, so if you merge/rebase, it should be green ;)

martpie:
This is actually the correct solution, the data management is a bit messy I got to say, I should rework that a little at some point.

Issue #514, opened 2020-02-16
Bug: Song duration not detected + infinite progress bar


qcasey:
I want to thank you for keeping this repo up to date!

On a very limited number of songs the duration is not detected, and the progress bar when playing these songs scales infinitely. I can look into both these issues since I have the files but I wanted to keep a tab of it here for easier merging.

![Screenshot from 2020-02-16 14-49-04](https://user-images.githubusercontent.com/1429672/74614737-5dc70d80-50cf-11ea-8cec-b4c4d74d4ae4.png)

martpie:
Gotta say I laughed at the screenshot 😄 thank you for reporting these and submitting fixes 🙌

PR #515, opened 2020-02-17
Issue #514

qcasey:
This fixes the 2 issues I mentioned in #514 - was easier than I expected.

PlayingBarInfo.tsx: hides the (blue) playing bar if the elapsed time has scaled past 1. Ideally this never happens of course.

utils.ts: music-metadata determined several songs to have inaccurately short durations (0.042666666666666665s) not picked up by `!metadata.duration` . On this event default to secondary audio duration retrieval. Less than half a second seems reasonable for a double check?


martpie:
I'll test it later but it looks good 🙌

PR #517, opened 2020-02-18
max height is full window, -header and -footer

qcasey:
Getting real close to my perfect media player!

The problem is when playlists go beyond ~14, the footer is knocked way down the page & prevents scrolling.
Screenshot from 2020-02-18 13-10-28

Giving the main content 100% window height minus the header+footer height worked for me, but open to your testing and suggestions.

Here are screenshots at different viewpoints (all hidpi)

Screenshot from 2020-02-18 13-16-53

Screenshot from 2020-02-18 13-13-44

Screenshot from 2020-02-18 13-14-28

### DIFF
--- a/src/ui/App.css
+++ b/src/ui/App.css
@@ -10,6 +10,7 @@
   position: relative;
   display: flex;
   flex-direction: column;
+  max-height: calc(100vh - 50px - 30px);
 }
 
 .content {


martpie:
This is definitely a bug, not sure this is how it should be fixed, I'll check that tomorrow :)

qcasey:
No worries, thanks! :)

src/ui/App.css
```diff
@@ -10,6 +10,7 @@
   position: relative;
   display: flex;
   flex-direction: column;
+  max-height: calc(100vh - 50px - 30px);
```
vprigent:
```suggestion
  max-height: calc(100vh - 80px);
```

should give the same result ?
assuming we know the size of the footer through a variable, it could be good to use that one here.

qcasey:
Agreed, I just wanted to demonstrate where each of the 80px is coming from.
Ideally this would be variable so things don't break if the footer or
header changes size (from font enlarge, etc)

martpie:
I knew I knew this bug. The problem is solved in 4fa45f0e20e60ac63948046a9eaab58ddbb593d6

for the reference: https://stackoverflow.com/questions/33605552/how-to-prevent-a-flex-item-height-to-overflow-due-to-its-content

I'll try to draft a new release tomorrow.

qcasey:
Nice, thanks!

Issue #127, opened 2016-08-02
Equalizer


KeitIG:
With the Web Audio API, it would be possible to implement an equalizer (60, 170, 310, 600, 1k, 3k, 6k, 12k, 14k, 16k for example).

KeitIG:
http://superuser.com/questions/695843/how-to-use-equalizer-on-ffmpeg#696187

YurySolovyov:
https://github.com/fluent-ffmpeg/node-fluent-ffmpeg#audiofiltersfilter-add-custom-audio-filters

YurySolovyov:
I actually thought that we'll implement equalizer on WebAudio Layer, cause to apply it to ffmpeg output, you have to restart the process.

KeitIG:
It was just for the record.

OutsourcedGuru:
[Some cool examples here](https://dzone.com/articles/exploring-html5-web-audio)

martpie:
Years later. I don't think Museeks should be in charge of providing an EQ, this should be the system or peripheral's drivers job.

PR #521, opened 2020-03-15
Introduce Gapless Playback

### LINKED ISSUE #31: Try to reduce sound interuption between two songs
Currently, when a song is over, the `src` of the audio tag is changed, but the tag has to preload the file a bit. So there is a micro interruption of a few miliseconds.

Could be fixed using a second audio tag.

Test needed.

### LINKED ISSUE #128: Switch from Audio to AudioContext to play music
The `Audio` media element works well, but we need the power of `AudioContext` to add effects, play without interruptions, easy cross fade ect.

That's a huge work, and here are the steps

- [ ] Create a `Player` class: abstract all audio manipulation to a class instead of manipulating the `HTMLMediaElement` directly
- [ ] Switch from `Audio` to `AudioContext` and use the Web Audio API

This issue will naturally lead to #127 #31  #114

s0:
This PR introduces gapless playback to museeks!

The core of the functionality is actually in an [external library](https://www.npmjs.com/package/@synesthesia-project/precise-audio) that I put together (mostly for the purposes of getting gapless playback to work in museeks).

The result is a somewhat less intrusive PR, most of the changes are in `src/ui/reducers/player.ts` to ensure that whenever the player or queue state changes, the `PreciseAudio` instance is updated with the expected queue.

This now also introduces an `AudioContext` and audio graph into museeks, which means that we can go ahead and implement #127 and other stuff after this!

Fixes #128
Fixes #31

I've tested this mostly on LAME-Encoded MP3s, but there are many more codecs that need to be supported, the good news is that now that the boilerplate is in, adding new codecs is simply a matter of adding them in [`@synesthesia-project/gapless-meta`](https://www.npmjs.com/package/@synesthesia-project/gapless-meta), and won't require any more plumbing. If you have any examples of gapless audio files that don't work, please let me know.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -252,6 +252,20 @@
       "integrity": "sha512-9NET910DNaIPngYnLLPeg+Ogzqsi9uM4mSboU5y6p8S5DzMTVEsJZrawi+BoDNUVBa2DhJqQYUFvMDfgU062LQ==",
       "dev": true
     },
+    "@synesthesia-project/gapless-meta": {
+      "version": "0.0.1",
+      "resolved": "https://registry.npmjs.org/@synesthesia-project/gapless-meta/-/gapless-meta-0.0.1.tgz",
+      "integrity": "sha512-WDldxaJ3E5ueR+0GW2MLafnXLXLDOClny/venpMpbyAkHSSJw8N9oScy69bUfmNp9poRmxsbNYF0xSgrWcAPLA=="
+    },
+    "@synesthesia-project/precise-audio": {
+      "version": "0.2.3",
+      "resolved": "https://registry.npmjs.org/@synesthesia-project/precise-audio/-/precise-audio-0.2.3.tgz",
+      "integrity": "sha512-oB4JWsMzjEN+AT7IEvOxFMnthlO4MPfJn2onwwlUxtR+K8Nj1scXpmYtNUejMdFSXg8fvfGzvJpkLlcTaSr+RA==",
+      "requires": {
+        "@synesthesia-project/gapless-meta": "^0.0.1",
+        "soundbank-pitch-shift": "~1.0.3"
+      }
+    },
     "@szmarczak/http-timer": {
       "version": "1.1.2",
       "resolved": "https://registry.npmjs.org/@szmarczak/http-timer/-/http-timer-1.1.2.tgz",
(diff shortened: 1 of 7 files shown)


martpie:
A lot of things (but actually less than I expected), thank you very much for the PR, I will try to have a look at it when I can :)

qcasey:
Thanks for the PR, this is very interesting. 

On my machine the memory usage at idle is about the same, which is nice.

Unfortunately loading songs from disk takes a lot longer and a lot more memory than master. 3 min songs take about 3 seconds, while 2+ hour audiobooks take ~20 seconds. The UI seems to hang in the meantime. When the audiobook is finally playing Museeks takes about 2 gigs of memory.

I know #128 in particular is a big change and I will try to help where I can. I like your idea of configuring a threshold for memory usage, but I don't think that will solve load times. 

Are you experiencing the same? Or is it time for me to upgrade my SSD :smile: ?

s0:
> Thanks for the PR, this is very interesting.
> 
> On my machine the memory usage at idle is about the same, which is nice.
> 
> Unfortunately loading songs from disk takes a lot longer and a lot more memory than master. 3 min songs take about 3 seconds, while 2+ hour audiobooks take ~20 seconds. The UI seems to hang in the meantime. When the audiobook is finally playing Museeks takes about 2 gigs of memory.
> 
> I know #128 in particular is a big change and I will try to help where I can. I like your idea of configuring a threshold for memory usage, but I don't think that will solve load times.

It should solve the load times too for the longer tracks, as the `HTMLAudioElement` allows us to inspect the duration before having loaded the entire file. So if the threshold was somewhere between 3 mins and 2 hours, for your case above, you'd experience a 3 second load delay for the song, and no delay for the audiobook (as it wouldn't try to load all of it).

> Are you experiencing the same? Or is it time for me to upgrade my SSD  ?

Memory yes, but those load times are significantly longer than I've experienced! Switching to an SSD isn't a bad idea ;)

I'll start working on the threshold idea now.

s0:
@qcasey I've updated the library with the threshold capability, which uses an `HTMLAudioElement` instead of an `AudioBuffer` for longer tracks, and updated this PR to use that.

I've currently configured it with a threshold of 10 minutes, so anything below that will play gaplessly, and anything above will use the old method, I'm thinking it probably makes sense to make that a user-configurable option in this PR?

Could you test these updates with the same files you used earlier?

qcasey:
Hope everyone is well!

No unfortunately I don't notice a difference, in either >10 or <10 min songs

s0:
> Hope everyone is well!
> 
> No unfortunately I don't notice a difference, in either >10 or <10 min songs

Hi @qcasey, sorry for the late reply, it's been a very few hectic months for me! :joy:

Have you got an example track that we could try, i'm surprised that you didn't see a difference with RAM usage.

I'm thinking about perhaps another alternative where we decode the whole song, but only keep a chunk of that in memory and drop the rest, and when we get near to the end of the chunk, we decode the song again and keep the next chunk. This would probably be something quite long (e.g. 10min chunks). This would keep RAM usage down on average, but every ~10 mins, we'd need a bunch of ram to decode the song and then drop it again.

This would be improved if we could find a way to decode only a portion of a song and not require the whole thing be in RAM, but unfortunately the current audio APIs don't provide a way of doing this. Perhaps we could call out to something else (like ffmpeg) for the decoding part? Though this isn't something iv'e done before.

PR #531, opened 2020-05-01
Normalize backslashes in track paths to fix broken playlists on Windows

GilgameshxZero:
On Windows, museeks@0.11.4 adds tracks with slash paths ("C:/path/track.mp3"), but interprets playlist tracks with backslash paths ("C:\path\track.mp3"). Without resolving track paths (thus normalizing slashes to backslashes), playlists will never find tracks to add, and thus will always be empty.

You can see this by creating a test directory, and placing a `track.mp3` and a `playlist.m3u` in the directory. In `playlist.m3u`, have one line adding the track:

```
track.mp3
```

Upon adding the directory to the library on museeks@0.11.4, the playlist will be empty (here, `resolved filePath` is without the resolution; I just forgot to take out the `console.log`):

image

However, by resolving the `filePath` for each track, the playlist will be interpreted correctly:

image

The PR also updates a small number of outdated comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Please consider that **`master` is unstable**.
 - `git clone git@github.com:martpie/museeks.git`
 - `cd museeks`
 - `npm ci`
-- `npm modules:rebuild`
+- `npm run modules:rebuild`
 - `npm run build` or `npm run dev`
 - `npm run museeks` or `npm run museeks:debug`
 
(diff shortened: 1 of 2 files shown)


martpie:
Good catch, thank you very much!

martpie:
I will try to release this tomorrow 👍

PR #549, opened 2020-10-30
Calculate PlaylistsNav height and scroll correctly

GilgameshxZero:
Currently, the playlist navigation miscalculates its height and doesn't invoke the scrollbar sometimes and hides the last playlist entry.
image
This fixes the height calculation in CSS.
image


martpie:
Thank you!

martpie (approved):

Issue #575, opened 2021-02-27
Audio seek using left/right arrow keys


mike-seger:
Hi Pierre

How about an audio seek as suggested here:

https://github.com/mike-seger/museeks/tree/audio-seek

Thanks a lot for your very nice and lean player.

Cheers
Mike

jvegaf:
Congrats ! It's a great feat. Why you don't make a PR of that branch ?  

Regards

jvegaf:
based in your idea, I implemented this feature using the keybinding component. 

https://github.com/jvegaf/museeks/commit/15c95b13780d3420e19b1fd25def17e464ecfc32

I'll try a make a PR.

martpie:
Thank you @jvegaf ✅

PR #588, opened 2021-06-14
+/- 10s audio seek with Arrow Keys

jvegaf:
Based on #575 

Seek +/- 10 sec with Arrow Keys


mike-seger (approved):

martpie:
@jvegaf thank you for the PR, sorry for the big delay for the review.

Overall, it looks good, but I think you should add a safeguard for when we are typing in an input or something:

Something like there: https://github.com/martpie/react-keybinding-component/blob/6688095b81dfe16c29798cc00e82ed28942e0a02/src/keybinding.tsx#L50

So maybe you can make use of the `<Keybinding />` component instead and pass it `preventInputConflict` :)

martpie:
Oh my bad, it's already the case, let's go then!

PR #590, opened 2021-06-18
Update metadata from native parser

HaiDang666:
I found that some metadata do not sit in `common` but in `native` from `music-metadata` package.

Screen Shot 2021-06-17 at 16 46 38

Before:
Screen Shot 2021-06-18 at 13 26 43

After:
Screen Shot 2021-06-18 at 13 02 04


martpie:
Can you send me some of those files per email so I can test them myself?

HaiDang666:
> Can you send me some of those files per email so I can test them myself?

sure but do we have a discorg server for this project ?

martpie:
We used to have a Gitter, but maybe a Discord could be a good idea indeed.

Thanks for the files!

HaiDang666:
Thank. It will be convenient to communicate.

HaiDang666:
should I close this since the metadata package has done it?

PR #593, opened 2021-07-26
Start work on electron.remote deprecation

martpie:


martpie:
Moving menus to the main process is going to take a longer time and probably need more back-end to be moved there first (DB management, etc)

PR #594, opened 2021-07-27
Portable win32 binaries

### LINKED ISSUE #587: Windows portable version
I would suggest making a portable version for Windows. From what I understand Museeks AppImage for Linux holds the default AppImage ruleset and hence is automaticaly portable. For macOS one cannot really make a portable app. For Windows you certainly can and programs are supported with this release format.

martpie:
- Fixes #587
- Fixes ia32 windows binaries not created (🙈)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -174,4 +174,7 @@ jobs:
         with:
           name: binaries-windows
           path: |
-            build/museeks-setup.exe
+            build/museeks-ia32-setup.exe
+            build/museeks-x64-setup.exe
+            build/museeks-ia32-portable.exe
+            build/museeks-x64-portable.exe
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -54,7 +54,11 @@ win:
       arch:
         - x64
         - ia32
+    - target: portable
+      arch:
+        - x64
+        - ia32
   icon: src/images/logos/museeks.ico
-  artifactName: ${name}-setup.${ext}
-# nsis:
-#   perMachine: true
+  artifactName: ${name}-${arch}-setup.${ext}
+portable:
+  artifactName: ${name}-${arch}-portable.${ext}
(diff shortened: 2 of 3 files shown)


vivadavid:
Hi,

With the new changes introduced in 0.20.0, I can't see any more the portable version for Windows.

Will it be available in the future?

Thanks!

martpie:
@vivadavid yeaah, we kind of need an installer now because we need to ensure Webview2 is installed before running the app (which explains the 20x binary size reduction).

I will investigate though, on how to recreate portable binaries, but I don't think it will be easy.

vivadavid:
> @vivadavid yeaah, we kind of need an installer now because we need to ensure Webview2 is installed before running the app (which explains the 20x binary size reduction).
> 
> I will investigate though, on how to recreate portable binaries, but I don't think it will be easy.

If it's a big deal, it doesn't matter: I'm just using the installer. However, in the long term, it'd be useful to have a portable version.

Thanks for your quick reply.

martpie:
Can you explain more in detail why a portable version is more desirable than a classic installer? 

Is it just the ease of use? A lack of admin right? Something else?

vivadavid:
In my particular case, when I learn about new software and I want to test it to know if it's of any use to me, I prefer not to write anything into the registry. Also, certain uninstallers don't work perfectly well and you end up keeping undesired files on your computer.

There's another case scenario. If I have, for example, Subtitle Edit in portable version, I can download all the Whisper models that I need and keep them all in the same folder. If, later on, I need to format my PC and have a fresh Windows install, I don't have to install the programme againt, go through all the settings, download the Whisper models, download Tesseract, etc.

In the case of a music player, I guess that, the first time your music folder is scanned, a library is generated, possibly even with a list of all your covers. It's a clean solution to have both the programme and the music library in the same folder, so that, if you format your PC, you don't have create your library from scratch again.

PR #610, opened 2021-09-30
Make utility

jvegaf:
*  Now can run yarn scripts more easily and quickly, calling **make** 🛠️  command

* Examples:

| BEFORE | AFTER |
| ------------ | ---------- |
| yarn install --frozen-lockfile | make |
| yarn run build | make build |
| yarn run dev | make dev | 
| yarn run museeks:debug | make debug | 

* You can run a determinate test:

| BEFORE | AFTER |
| ------------ | ---------- |
| yarn run test:unit | make test/unit | 

* Or run **ALL TESTS** and make a build with:

```shell script
make test
```
* You can remove **node-modules** folder  with

```shell script
make save-the-world
```

😆


martpie:
Hello 👋

very sorry for the late response and thank you for this.

To me it is a non-issue as the following is correct too:

| BEFORE | AFTER | ALSO |
| ------------ | ---------- | - | 
| yarn install --frozen-lockfile | make |  yarn |
| yarn run build | make build | yarn build |
| yarn run dev | make dev |  yarn dev |
| yarn run museeks:debug | make debug | yarn museeks:debug |

So I see very little added value in having a makefile, especially if it's linux-only. An update to the README to add the simplified command could be a nice thing though.

PR #667, opened 2022-12-25
DB from Main Process

martpie:


martpie:
Not exactly what was planned, but still nice improvements.

PR #671, opened 2023-03-20
Better built executable naming

zyrouge:
- Adds platform infix like `win`, `linux` and `macos`.
- Added `tar.gz` target in linux.


martpie:
That actually makes a lot of sense, thank you!

martpie (approved):

PR #683, opened 2023-06-14
Feat: Make click element cursor pointer

dipras:
Just a small contribution for make button cursor pointer :)

### DIFF
--- a/src/renderer/components/PlayerControls/PlayerControls.module.css
+++ b/src/renderer/components/PlayerControls/PlayerControls.module.css
@@ -11,6 +11,7 @@
     font-size: 16px;
     padding: 0 10px;
     color: inherit;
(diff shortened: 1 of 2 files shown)


martpie:
Hey, thank you for this!

This was actually a deliberate choice following my read of: https://adamsilver.io/blog/buttons-shouldnt-have-a-hand-cursor/ and to follow defaults.

That said, I've heard some major browsers are changing this default to be "pointer", so I have to do some additional research. Let's keep this open in the meantime :)

martpie:
Current design system is not great and need to be re-designed. For example, the difference between inputs and buttons is not clear enough. For this reason, adding a hand cursor in the meantime is a good solution, thanks!

martpie (approved):

PR #690, opened 2023-06-28
Config from Main Process only, drop teeny-conf, use electron-store

martpie:


martpie:
Anyway, that's needed in order to move all usage of Node.js from the Renderer process, so even if there's some hacky stuff, it's ok, and still a win.

PR #694, opened 2023-06-29
Header improvements

martpie:
A bunch of stuff, but basically: 

- Bigger horizontal space for tracks metadata
- Title & Artists + Album on two line

Still to think about:
- Re-think volume?
- Fix the muted weird behavior?
- Not satisfied with progress text `(00:00 / 00:00)`

### Before 
Screenshot 2023-06-29 at 18 49 35

### After 

Screenshot 2023-06-29 at 18 50 32

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@electron/remote": "^2.0.10",
+    "@radix-ui/react-aspect-ratio": "^1.0.3",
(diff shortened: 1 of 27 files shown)


dipras:
I think the volume control must be triggered if the volume icon clicked. Because sometimes when i want to interact with music control(previous, pause, next), i accidentally triggered the volume control

PR #696, opened 2023-07-03
DB 3.0

martpie:
LinvoDB has been dead for 5+ years. I am experimenting with new DBs. 

Ideally, using SQLite + an ORM would be awesome, but distributing SQLite is a surprisingly undocumented problem. PouchDB or @seald/nedb are the simplest options.

_edit:_ I went with PouchDB, this is a breaking change and will require people to export their playlists first.


martpie:
Early experiments with PouchDB, and wtfff?

PR #700, opened 2023-07-06
Feat: Scan theme directory

dipras:
As requested in src/shared/lib/themes.ts to improve scaning json file instead import one by one
I make the app auto scan all json in themes folder to be theme option for museeks. I used `MuseeksAPI` as bridge to deliver scan result to frontend because file system cannot be scanned from frontend


package-lock.json
martpie, line 1: Please remove this file :) we use yarn

yarn.lock
martpie, line 1: as you did not add any dependency, this file should not have been updated. Can you reset it? (running `yarn` should be enough)

(I use yarn v1, I'll migrate to v3 + pnp at some point though)

src/shared/lib/themes.ts
```diff
@@ -1,11 +1,14 @@
...
-import darkLegacyTheme from '../themes/dark-legacy.json';
+import fs from 'fs';
 
-export const themes = [
-  lightTheme as Theme,
-  darkTheme as Theme,
-  darkLegacyTheme as Theme,
-];
+const themes : Array<Theme> = [];
+fs.readdirSync('./src/shared/themes').forEach(name => {
```
martpie, line 5: I am not 100% sure it would when the app is packaged. 

Can you try two things?

- `yarn build && yarn museeks`
- `yarn build && yarn package:l` (if you use linux)
  - try to install whatever package you are interested in (from `build/`)
  - run the app and see if something wrong happens

martpie (changes requested):
Thanks for that! That could be a great base for custom themes, as we could imagine scanning a custom folder, like `~/.museeks/themes` on startup, and users could put their themes there.

Of course that would mean we should do a couple of runtime validations, to make sure the keys are aligned with the latest theme definition, and we should fallback to `_system` in case a theme is invalid.

But I think we can discuss it in a later diff :)

martpie:
Also, Q&A is failing, I should probably add the list of required VSCode extension required to make things work.

dipras:
> Thanks for that! That could be a great base for custom themes, as we could imagine scanning a custom folder, like `~/.museeks/themes` on startup, and users could put their themes there.
> 
> Of course that would mean we should do a couple of runtime validations, to make sure the keys are aligned with the latest theme definition, and we should fallback to `_system` in case a theme is invalid.
> 
> But I think we can discuss it in a later diff :)

Ah yep, that's great idea we can make an custom theme for user. We can do it later after this PR merged

package-lock.json
dipras: Okay, i will remove this file

src/shared/lib/themes.ts
dipras: I Tried to package the app and i found there is problem with the build and  i found it in master branch too
image

At first i thought maybe the missing modules is listed in `devDependency` instead in `dependency`, so i tried to look into package.json and yep that is not the problem because it listed in `dependency`
The app told pino module is missing. I tried to remove pino from the code, another module is missing too
image

So, the problem is the builder and i don't know it's only in linux or other device.

dipras: Because this issue have solved, then i will fix the rest

martpie:
So I just tested it, there are two issues: 

- we do not control the order of the order of the themes anymore, which is a problem:

Screenshot 2023-07-10 at 01 55 09

- packaged apps crash:

Screenshot 2023-07-10 at 01 53 17

As previously said, I don't think it's an important problem to be solved, but we can keep part of this code! My suggestion would be to keep the old behavior, but use this PR as a base for custom themes: could you do the same, but by scanning `~/.config/museeks/themes`? Because it's an absolute path, things should not crash.

You can use this API to get the right path: https://www.electronjs.org/docs/latest/api/app#appgetpathname

No need to do fancy checks right now, you can just steal one of the theme, change its name, update two variables and test it.

wdyt?

martpie (changes requested):
as-per-my-previous-comment

dipras:
> So I just tested it, there are two issues:
> 
> * we do not control the order of the order of the themes anymore, which is a problem:
> 
> Screenshot 2023-07-10 at 01 55 09
> * packaged apps crash:
> 
> Screenshot 2023-07-10 at 01 53 17
> As previously said, I don't think it's an important problem to be solved, but we can keep part of this code! My suggestion would be to keep the old behavior, but use this PR as a base for custom themes: could you do the same, but by scanning `~/.config/museeks/themes`? Because it's an absolute path, things should not crash.
> 
> You can use this API to get the right path: https://www.electronjs.org/docs/latest/api/app#appgetpathname. `appData` seems the right location, but it's not easily reachable for Windows/Mac, but don't worry, I'll think about a solution. Once the solution works, we can easily change the path to a better one.
> 
> No need to do fancy checks right now, you can just steal one of the theme, change its name, update two variables and test it.
> 
> wdyt?

Okay, i think it's a good idea. i will try to implemant it

Issue #701, opened 2023-07-07
Module not found when running build package


dipras:
I Tried to package the app and it's success, also when install the build on my device there is no problem at all. But when i run the app, it won't open. So i thought this is the problem in my code to add this feature. I tried to change the branch to master but the result is same, the app won't open. So i tried to execute it with console and i found there is problem with the build
image
![image](https://github.com/martpie/museeks/assets/51155424/0a62e1ab-66a5-4512-8c91-fa6cac82eab3)

At first i thought maybe the missing modules is listed in devDependency instead in dependency, so i tried to look into package.json and yep that is not the problem because it listed in dependency
The app told pino module is missing. I tried to remove pino from the code, another module is missing too
image
![image](https://github.com/martpie/museeks/assets/51155424/c792a3c7-23e6-408a-b8e4-70d311bf5480)

dipras:
Okay it's seem i found the problem, the problem is in here
https://github.com/martpie/museeks/blob/c040a1ab488478a817429545ad04a0c111a8de20/electron-builder.yml#L5-L8

On line 8, all node files is exclude to package the build. So the package is not included and the package face error about missing module

martpie:
Can you send reproduction steps? Nothing from `node_modules` should be included in the repo.

dipras:
https://github.com/martpie/museeks/assets/51155424/21c77bf3-3a12-4c7c-bbd0-b5a3e066f4c1

Note: I change `electron-builder.yml` to build only one file in purpose for test and fast

martpie:
I can reproduce, investigating :)

PR #702, opened 2023-07-07
Fix: module not found in package

dipras:
An issue according to #701


martpie:
That's definitely not the right way to solve the issue, it's probably a bundling problem that I missed. Can you send reproduction steps on master?

martpie:
That is actually completely correct and I'm a moron.

PR #748, opened 2024-03-03
Tauri

martpie:
Tauri

This is a half-working copy of Museeks ported to Tauri 2.0.

This is highly experimental, many things are broken. Lots of non-idiomatic Rust (and React actually).

```bash
yarn
cargo test --manifest-path src-tauri/Cargo.toml
yarn tauri dev
```

PR #749, opened 2024-03-16
Capture logs while initializing plugins

uklotzde:


martpie:
Thanks for that, I was wondering why logging was not working in plugins.

Let me create an official tauri branch for all that, and we can start merging things there (I use stacked diffs for #748, so I'm basically force-pushing all the time).

The branch is still missing a few critical pieces, but I can create new PRs for these

martpie:
(bis) Due to force-pushes, I think there are conflicts here. Could you maybe just rebase on top `tauri` and ignore [aa1b762](https://github.com/martpie/museeks/pull/750/commits/aa1b7627f7e80e7c43d4d5c46ba7eabb38d340e0) ?

Sorry, that's due to me using Sapling SCM for managing my work, and it does not work well in a collaborative environment with GitHub.

martpie:
Thank you! I am not a Rust person, so you will see a lot of non-idiomatic Rust in the current wip codebase for Tauri ;)

PR #753, opened 2024-03-17
Tauri

martpie:
```bash
yarn
cargo test --manifest-path src-tauri/Cargo.toml
yarn tauri dev
```

Let me keep track of what's working and what's yet to be fixed using the `tauri` branch. Non-exhaustive. Hard bockers are marked with 🚫.

- [x] Database setup **(breaking change)**
  - [x] Prevent importing tracks already in library
- [x] Config setup **(breaking change)**
- [x] Track import
- [x] Loading UI
  - [x] Fix all TS issues
- [x] Header
  - [x] macOS: overlayed window controls
  - [x] Player controls
  - [x] Search
    - [x] Fix select on focus
  - [x] Hover on progress border moves the tooltip to the wrong location
- [x] Footer
  - [x] Progress tracking
- [x] Library
  - [ ] Smart scan (pick multiple folder and/or tracks and/or playlists)
  - [x] Playback
    - [ ] start button with no queue does not work anymore
  - [x] Context Menu
  - [x] Delete tracks
  - [ ] Drop files on the window to import them
    - [ ] Windows, can it work? https://github.com/tauri-apps/wry/issues/904
  - [x] No way to "show in file manager" https://github.com/tauri-apps/plugins-workspace/issues/999
- [x] Playlists
  - [x] Auto-import m3u
  - [x] Export to m3u
  - [x] Drag-and-Drop effect is cut when "below"
  - [x] CRUD
    - [x] Create
    - [x] Edit
    - [x] Delete
    - [x] Reorder
      - [x] Fix `get_multiple` not returning docs in the right order
  - [x] Playback
- [x] Covers
  - [x] ID3 -> Base 64
  - [x] Scan folder for `cover.jpg`, etc
- [ ] Settings / Integrations
  - [ ] 🚫 Theme
    - [ ] Dynamically set window theme at runtime https://github.com/tauri-apps/tauri/issues/5279
    - [ ] `__system`  theme
  - [ ] 🚫 Audio output selection
    - [ ] What of macOS? seems impossible https://bugs.webkit.org/show_bug.cgi?id=216641
  - [x] Single instance
  - [x] macOS close/activate without destroying the window https://github.com/tauri-apps/tauri/issues/3084
  - [x] Devtools
  - [x] Dialogs
  - [ ] Dock actions (macOS)
  - [x] Menu
    - [x] 🚫 Menu Events
      - [x] Go to settings
      - [x] Go to library
      - [x] Go to playlists
      - [x] Jump to playing track
      - [ ] Accelerators are broken on Windows https://github.com/tauri-apps/wry/issues/451
    - [x] Show/Hide menu on Alt in Windows/Linux
  - [x] Sleep Blocker
  - [ ] Thumbbar
  - [x] Window Position
  - [x] Sleepblocker
  - [x] Track change notifications
- [x] CI
  - [x] Pass QA
  - [x] Build binaries
    - [x] Windows
    - [x] macOS
    - [x] Linux
- [ ] Misc
  - [x] Listen events on macOS is broken? https://github.com/tauri-apps/tauri/issues/9052
  - [ ] TS exports to a single file https://github.com/Aleph-Alpha/ts-rs/issues/59
  - [x] `Unhandled Promise Rejection: QuotaExceededError: The quota has been exceeded.` on webkit
  - [ ] Hot reload breaks routing refresh
- [ ] Accessibility 
  - [ ] Bring back focus ring
- [ ] Final test
  - [ ] Windows
  - [ ] macOS
  - [ ] Linux

## Screenshots

Screenshot 2024-03-05 223247

Screenshot 2024-03-12 at 16 31 17


Serial-ATA:
Hey @uklotzde & @martpie , I fixed that bug in WebKit and enabled MediaSession by default. The player should work fine on Linux now whenever my commits make it into a release (I'm still not entirely sure about how their release schedule works).

uklotzde:
Generating TypeScript bindings is a burden.

Those definitions are probably better defined once as part of the application API in museeks. Every integration then needs to implement the bridging code manually instead of bringing their own definitions.

uklotzde:
> Edit: Typically, by running `cargo test` in the `beforeDevCommand`. But that would run cargo twice (test + build), but maybe things are cached.

Only a single TypeScript API and those definitions should be driven by the requirements and use cases of the app and rendering. If you want pluggable backends then backends need to adapt these definitions instead of bringing their own.

I already have JSON representations and  JSON schema mapping and started to utilize `ts-rs`. But this turns out as complicated and wouldn't really help if you need a second layer of type adapters for app integration anyway. Mapping the idiomatic, native Rust types directly instead of opinionated, JSON-serializable types makes more sense here.

martpie:
I think I don't fully understand what you mean.

Can you detail how using JSON schemas would be an improvement and how the dev workflow would look like?

Serial-ATA:
> 🚫 Audio output selection
>
>    What of macOS? seems impossible [bugs.webkit.org/show_bug.cgi?id=216641](https://>bugs.webkit.org/show_bug.cgi?id=216641)

Just letting you know, it doesn't seem to work on *any* platform, not just macOS. WebKit just returns a list of inputs.

martpie:
Yep, that is what I figured out too, testing on Windows only produces a list of audio/video inputs when request permission with `getUserMedia()`, audio outputs seem to just not be working. 

To be completely frank, I think I will just remove the feature, people should probably rely on system settings to route their app to specific audio outputs ([example for windows](https://winaero.com/audio-output-device-apps-individually-windows-10/), for windows and linux, good luck).

https://github.com/martpie/museeks/issues/449 was the original feature request. 

I should also check how other players deal with it (app setting vs system setting)

martpie:
Actually nevermind, this seems to be working on Windows after calling

```
await navigator.mediaDevices.getUserMedia({
  audio: true,
});
```

`setSinkId` works too.

It does require giving access to the microphone though, which is privacy-sensitive and users make be scared of seeing this popup.

image


 image

Serial-ATA:
> I should also check how other players deal with it (app setting vs system setting)

Being able to choose an output is a pretty common feature, but they'll use the system default initially, of course.

> people should probably rely on system settings to route their app to specific audio outputs

Yeah, it's trivial anyway to use something like [pavucontrol](https://flathub.org/apps/org.pulseaudio.pavucontrol) (on Linux) to choose outputs.

Serial-ATA:
I can look into getting the outputs to show up on Linux at least. Just looks like a case similar to the MediaSession bug where they lost interest and left the issue sitting there.

martpie:
damn, stop being awesome matey

uklotzde:
Config broken after updates?

```
yarn run v1.22.21
$ tauri dev
       Error `tauri.conf.json` error on `build`: Additional properties are not allowed ('devUrl', 'frontendDist' were unexpected)
       Error `tauri.conf.json` error: Additional properties are not allowed ('app', 'bundle', 'identifier', 'productName', 'version' were unexpected)
error Command failed with exit code 1.
```

martpie:
Did you run `yarn` before `yarn tauri dev`?

uklotzde:
> Did you run `yarn` before `yarn tauri dev`?

Thanks, that fixed it. Not familiar with Node.js tooling :grimacing:

martpie:
Let's merge this. It's not ready, but it will show the project is alive :)

PR #755, opened 2024-03-17
Replace home-config with directories+toml

uklotzde:
Instead of creating config/data directories manually the [directories](https://crates.io/crates/directories) crate should be used for platform-independent standard locations.

On Linux the files land in `{HOME}/.config/io.museeks.app/` (local config, also used by Tauri) and `{HOME}/.local/share/io.museeks.app/` (local data).

Please adjust as desired.

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -27,7 +27,7 @@ base64 = "0.22.0"
 bonsaidb = { version = "0.5.0", features = ["local", "async"] }
 dirs = "5.0.1"
 futures = "0.3.30"
-home-config = { version = "0.6.0", features = ["toml"] }
+directories = "5.0.1"
 log = "0.4.21"
 lofty = "0.18.2"
 memoize = "0.4.2"
@@ -38,6 +38,7 @@ serde = { version = "1.0.197", features = ["derive"] }
 strum = { version = "0.26.2", features = ["derive"] }
 thiserror = "1.0.58"
 tokio = { version = "1.36.0", features = ["macros"] }
+toml = { version = "0.8.11", default-features = false, features = ["parse"] }
 ts-rs = "7.1.1"
 uuid = { version = "1.7.0", features = ["v3", "v4", "fast-rng"] }
 walkdir = "2.5.0"
--- a/src-tauri/src/libs/utils.rs
+++ b/src-tauri/src/libs/utils.rs
@@ -1,8 +1,14 @@
+use directories::ProjectDirs;
 /**
  * Small utility to display time metrics with a log message
  */
 use log::info;
-use std::{ffi::OsStr, path::PathBuf, time::Instant};
+use std::{
+    ffi::OsStr,
+    fs,
+    path::{Path, PathBuf},
+    time::Instant,
+};
 use tauri::{Runtime, WebviewWindow};
 use walkdir::WalkDir;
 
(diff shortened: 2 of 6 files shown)


martpie:
yeah, this is one of the thing I know is wrong, but at the same time so much simpler haha.

I think it's for the best though, I will create a button in the settings to quickly open this folder (I have to clear it every once in a while while developing)

src-tauri/Cargo.toml
```diff
@@ -27,7 +27,7 @@ base64 = "0.22.0"
 bonsaidb = { version = "0.5.0", features = ["local", "async"] }
 dirs = "5.0.1"
 futures = "0.3.30"
-home-config = { version = "0.6.0", features = ["toml"] }
+directories = "5.0.1"
```
martpie, line 30: This should probably be changed in multiple ways:

- instead of `directories`, you can either re-use `dirs`, or better, Tauri's [`app_config_dir`](https://docs.rs/tauri/2.0.0-beta.11/tauri/path/struct.PathResolver.html#method.app_config_dir)

src-tauri/src/libs/utils.rs
```diff
@@ -28,12 +34,54 @@ impl TimeLogger {
...
- * Get the app configuration/storage directory
- */
-pub fn get_app_storage_dir() -> PathBuf {
-    let path = dirs::home_dir().expect("Get home dir");
-    path.join(".museeks")
+#[must_use]
+const fn app_name() -> &'static str {
+    // Must match the value of "identifier" in tauri.config.json!
+    "io.museeks.app"
+}
```
martpie, line 41: you should get that from Tauri: https://docs.rs/tauri/2.0.0-beta.11/tauri/struct.Config.html#structfield.identifier

src-tauri/src/libs/utils.rs
```diff
@@ -28,12 +34,54 @@ impl TimeLogger {
...
+    if let Err(err) = fs::create_dir_all(app_dir) {
+        log::error!(
+            "Failed to create app directory '{dir}': {err}",
+            dir = app_dir.display(),
+        );
+    }
+}
+
+#[must_use]
+fn init_config_dir(app_dirs: &ProjectDirs) -> &Path {
```
martpie, line 58: imho, all config related code should go to the `config.rs` for the sake of not jumping between 5 files when reading the config code.

src-tauri/src/libs/utils.rs
```diff
@@ -28,12 +34,54 @@ impl TimeLogger {
...
+        .map(init_config_dir)
+        .map(Path::to_path_buf)
+}
+
+#[must_use]
+pub fn app_data_dir() -> Option<PathBuf> {
+    app_dirs()
+        .as_ref()
+        .map(init_data_dir)
+        .map(Path::to_path_buf)
```
martpie, line 84: In the end, that's quite some additional code, so I would suggest to maybe keep `home_config` for the sake of simplicity, but just change the name of the directory for the config. (cf. pointer above to get the identifier + config path

martpie (changes requested):

src-tauri/Cargo.toml
uklotzde: `dirs` is a lower-level API and not useful.

Tauri already provides a `directories` API. But then initialization has to be deferred until `setup()` when an `AppHandle` is available.

PR #757, opened 2024-03-17
Use app_local_data_dir() for database

uklotzde:


src-tauri/src/plugins/database.rs
```diff
@@ -573,8 +576,18 @@ pub fn init<R: Runtime>(db: DB) -> TauriPlugin<R> {
...
+            tauri::async_runtime::spawn(async move {
+                let db = match setup(&app_handle).await {
+                    Ok(db) => db,
+                    Err(err) => {
+                        error!("Failed to setup database: {:?}", err);
+                        return;
+                    }
+                };
+                app_handle.manage(db);
+            });
```
martpie, line 590: very nice!

martpie (approved):
Thank you!

src-tauri/src/plugins/database.rs
uklotzde: Not sure about the timing implications. The plugin state might become available after the setup stage has already completed.

But it seems to work as expected, is supported according to the thread-safety constraints and I don't know of any alternative solutions.

martpie:
I don't really like that Tauri uses the bundle identifier as the path for configs/data, that's not as readable/findable as using the app name. I may change this behavior at some point, but let's stop using the home dir at least (though it's really convenient).

image

src-tauri/src/plugins/config.rs
```diff
@@ -136,6 +134,15 @@ pub fn set_config(config_manager: State<ConfigManager>, config: Config) {
     config_manager.update(config);
 }
 
+/**
+ * Get the app configuration/storage directory
+ */
+// TODO: Replace with PathResolver::app_config_dir().
+fn get_app_storage_dir() -> PathBuf {
+    let path = dirs::home_dir().expect("Get home dir");
```
martpie, line 142: So, how about using `dirs::config_dir();`? + `museeks` as folder.

uklotzde: `directories` would be more appropriate than the the low-level `dirs` crate.

But since this is supposed to be a temporary hack anything that respects the established standards on each platform is better than polluting the home directory.

PR #758, opened 2024-03-18
Update Tauri Deps

martpie:
I tried to upgrade `ts-rs` to v8, but the new export mechanism is extremely confusing, and `TS_RS_EXPORT_DIR` does not seem to be respected (types should be outputted in the TS codebase, not the rust codebase).


uklotzde:
Did you consider using [schemars](https://github.com/GREsau/schemars) to generate TypeScript from an intermediate JSON Schema definition?

martpie:
Well, ts-rs does everything in one go.

Probably we could automate the types generation when using `tauri dev`, I'll have a look.

PR #759, opened 2024-03-20
[Tauri][wip] Fix theme integration

martpie:
do not merge. not ready. code sucks. the whole process sucks. I have to draw a flow chart from scratch.

I need to move that code to a single-direction-data-flow.

What we need to reimplement:

- [ ] [Museeks 0.13 behavior](https://github.com/martpie/museeks/blob/277203176555331f88462d4ba2cf88d07d436ddc/src/main/modules/NativeThemeModule.ts#L102), but for Tauri
  - [x] "System" theme
    - [x] Window theme is correctly set (already done by Tauri, when `theme` is `None`)
      - [ ] **[hard]** While making explicit light/dark initial window theme work
    - [x] Update theme based on the `ThemeChanged` event
  - [ ] "Dark" or "Light" theme
    - [ ]  Set dark/light window theme on startup
    - [ ] Set dark/light app content theme on startup
    - [ ] **[hard]** Update window theme at runtime when user selects a new theme

### DIFF
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -14,8 +14,6 @@
         "label": "main",
         "title": "Museeks",
         "visible": false,
-        "hiddenTitle": true,
-        "titleBarStyle": "Overlay",
         "height": 550,
         "minHeight": 550,
         "width": 900,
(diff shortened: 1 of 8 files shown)


src-tauri/tauri.conf.json
```diff
@@ -14,8 +14,6 @@
         "label": "main",
         "title": "Museeks",
         "visible": false,
-        "hiddenTitle": true,
-        "titleBarStyle": "Overlay",
```
martpie, line 18: Remove that after testing

martpie:
Actually already fixed

PR #762, opened 2024-03-29
[Tauri] Working CI

martpie:
- [x] Fix QA
- [x] Build binaries


martpie:
noice. binaries are not uploaded yet, but let's maybe wait on https://github.com/tauri-apps/tauri-action/issues/342 

Otherwise, we'll use the upload-artifacts action to manually upload binaries.

PR #770, opened 2024-04-11
Biome

martpie:
Playing with Biome to see if it would be a good replacement for ESLint and Prettier. So far, I like it.

https://biomejs.dev


martpie:
lol 

Screenshot 2024-04-12 at 01 50 13

PR #790, opened 2024-09-28
Move some system theme logic back to front-end

martpie:


martpie:
This also fixes a bunch of issues on windows

Issue #732, opened 2023-10-05
Edit more id3 tags (art. date, etc)


kris20030705:
Hello dev!
i know this is minimal music player but please work in editing tags..
there is possible to edit tags but only artist,song,ablum & genre..
please add more option like: album art, lyrics. date.etc..
hope to see asap..

igorer88:
@martpie I would like to collaborate with that. Editing the tags on the file. There is a limitation? Or it wasn't  on the scope yet? Working with the lyrics would be fun too. Let me know where I can help.

martpie:
Thank you for willing to contribute! Here are some code pointers:

- Please read the README.md for the setup guide
- UI:
  - Editing more fields in the DB should be quite straightforward (and I just fixed it with the Tauri version of Museeks):
  - [`ViewTrackDetails.tsx`]([ViewTrackDetails.tsx](https://github.com/martpie/museeks/blob/master/src/views/ViewTrackDetails.tsx)) holds the form 
  - [`useLibraryStore`](https://github.com/martpie/museeks/blob/master/src/stores/useLibraryStore.ts#L176-L211) handles the back-end submission logic
  - [`database.ts`](https://github.com/martpie/museeks/blob/master/src/lib/database.ts#L23-L27) is a small wrapper around IPC communication (communication between the UI and the Back-End
- Back-end:
  - [`database.rs`] holds the logic updating the track. Right now it only overrides the database records, but does not set the ID3 tags.
  - In order to do so, you may need to leverage the rust library I use to scan the ID3 tags: [`lofty`](https://docs.rs/lofty/latest/lofty/). I believe they have methods to write ID3 tags.

I did not want to do it because it's a tricky feature and I really don't want Museeks for being responsible for destroying someone's library (hence the current "read-only" aspect of Museeks).

On a more UX side: we should probably make this ID3 write optional (and disabled by default). You can use a checkbox to ask the users what they want.

Good luck, and if you need any help, please ping me!

igorer88:
Nice @martpie I'm currently working on the form with the new fields in the Track type leaving outside _id.

Fixed a  minor bug on artists description hehe
<img width="613" alt="imagen" src="https://github.com/user-attachments/assets/2bc31940-2e01-4ba0-a3b6-d56cd2d2237b">

As far I have some doubts, thou.

1.   Inline inputs
[...]

PR #791, opened 2024-09-30
Feature/edit more id3 tags

igorer88:
This PR helps to solve #732

- Added some basic metadata missing fields to edit _track field_.
- Fix artist description label that was on title field.


martpie:
Ok, I'll start with your comments on #732:

- I fixed the DB imports, rebasing on `master` should make things work now

> The cover is present on the Other tag, while Museeks only searches on the Front Cover tag.

If we can scan for the cover in some kind of priority (`Front Cover` first, then `Other`, then then any cover), that would sounds good to me. 

> About storing the cover on the tag, well that's the standard way, it means that it would work on every other music player.

Ideally, it would be great to have the option to do both,. The reason I am a bit reticent for that is I am afraid of messing someone's library files, and that's why placing a `cover.jpg` in their folder seems safer, even though it's not x-player compatible. I am starting why some players copy all files to another location before changing them, but I would like to avoid doing that. 

Assuming a 512x512 should be <100KB, it's probably ok to embed them into the file itself. As the cover is not set in the DB anyway, I would do that separately from this PR though:

- 1 PR for more fields udpate
- 1 PR for optionally edit ID3 tags
- 1 PR for setting the cover as ID3

src/lib/utils-library.ts
```diff
@@ -120,3 +120,10 @@ const SORT_ORDERS: Record<SortBy, SortConfig> = {
 export function getSortOrder(sortBy: SortBy): SortConfig {
   return SORT_ORDERS[sortBy];
 }
+
+/**
+ *  Formats the track's duration from seconds to hh:mm:ss
+ */
+export function formatDuration(seconds: number) {
+  return new Date(seconds * 1000).toISOString().substring(11, 19);
+}
```
martpie, line 129: You can probably re-use that instead, as it's already in the codebase and is unit-tested:

https://github.com/martpie/museeks/blob/01ad926c002db3c11861ed8065aa20ca33b072cb/src/hooks/useFormattedDuration.ts#L3-L24

https://github.com/martpie/museeks/blob/01ad926c002db3c11861ed8065aa20ca33b072cb/src/hooks/__tests__/useFormattedDuration.test.ts#L5-L14

src/generated/typings/index.ts
```diff
@@ -30,5 +30,5 @@ export type SortOrder = "Asc" | "Dsc";
 /**
  * Track
  * represent a single track, id and path should be unique
- */
-export type Track = { _id: string, title: string, album: string, artists: Array<string>, genres: Array<string>, year: number | null, duration: number, track: NumberOf, disk: NumberOf, path: string, };
+*/
+export type Track = { _id: string, title: string, album: string, artists: Array<string>, genres: Array<string>, year: number, duration: number, track: NumberOf, disk: NumberOf, path: string, };
```
martpie, line 34: Why remove the `null` from track? This typing is generated via `cargo test` and mimics the definition in Rust:

https://github.com/martpie/museeks/blob/master/src-tauri/src/libs/track.rs#L18-L30

so if you don't keep them in sync the CI will scream at you.

src/views/ViewTrackDetails.tsx
```diff
@@ -26,12 +28,28 @@ export default function ViewTrackDetails() {
...
+    >
   >({
     title: track.title ?? '',
     artists: track.artists,
     album: track.album ?? '',
     genres: track.genres,
+    year: track.year ?? 0,
+    duration: track.duration,
+    track: track.track ?? { no: 0, of: 0 },
+    disk: track.disk ?? { no: 0, of: 0 },
```
martpie, line 51: I would be careful with those defaults. The real default should probably be `{ no: null, of: null }`. Also, `track.track` and `track.disk` should technically always be defined.

src/views/ViewTrackDetails.tsx
```diff
@@ -62,7 +80,6 @@ export default function ViewTrackDetails() {
         <Setting.Section>
           <Setting.Input
             label="Title"
-            description="You can add multiple artists with commas"
```
martpie, line 65: Good catch

src/views/ViewTrackDetails.tsx
```diff
@@ -115,6 +133,130 @@ export default function ViewTrackDetails() {
             }}
           />
         </Setting.Section>
+        <Setting.Section>
+          <Setting.Input
+            label="Year"
+            id="year"
+            name="year"
+            type="number"
+            min="1900"
```
martpie, line 142: You never know ;) classical music and all that. Min: 0 would probably be safer, even if highly improbable.

src/views/ViewTrackDetails.tsx
```diff
@@ -115,6 +133,130 @@ export default function ViewTrackDetails() {
...
+            step="1"
+            value={formData.year}
+            onChange={(e) => {
+              setFormData({ ...formData, year: Number(e.currentTarget.value) });
```
martpie, line 146: I think we need to be more careful here as `e` or `NaN` for example are considered a valid number in JS.

src/views/ViewTrackDetails.tsx
```diff
@@ -115,6 +133,130 @@ export default function ViewTrackDetails() {
...
+            }}
+          />
+        </Setting.Section>
+        <Setting.Section>
+          <Setting.Input
+            label="Duration"
+            id="duration"
+            name="duration"
+            type="text"
+            pattern="([01]?[0-9]{1}|2[0-3]{1}):[0-5]{1}[0-9]{1}"
```
martpie, line 156: Probably no need for a pattern here as it's never going to change. Maybe don't even include this field?

src/views/ViewTrackDetails.tsx
```diff
@@ -115,6 +133,130 @@ export default function ViewTrackDetails() {
...
+            readOnly={true}
+            value={formatDuration(formData.duration)}
+            onChange={(e) => {
+              setFormData({
+                ...formData,
+                duration: Number(e.currentTarget.value),
+              });
+            }}
```
martpie, line 164: `onChange` probably not needed as we are never going to edit this data.

src/views/ViewTrackDetails.tsx
```diff
@@ -115,6 +133,130 @@ export default function ViewTrackDetails() {
...
+          />
+        </Setting.Section>
+        <Setting.Section>
+          <Setting.Input
+            label="Duration"
+            id="duration"
+            name="duration"
+            type="text"
+            pattern="([01]?[0-9]{1}|2[0-3]{1}):[0-5]{1}[0-9]{1}"
+            readOnly={true}
```
martpie, line 157: I would suggest `disabled`, that way it won't receive focus, won't be send to the form event and will get appropriate styling

src/views/ViewTrackDetails.tsx
```diff
@@ -26,12 +28,28 @@ export default function ViewTrackDetails() {
...
+      | 'track'
+      | 'disk'
+      | 'path'
+    >
   >({
     title: track.title ?? '',
     artists: track.artists,
     album: track.album ?? '',
     genres: track.genres,
+    year: track.year ?? 0,
```
martpie, line 48: `null` by default please, those fields can be empty.

src/views/ViewTrackDetails.tsx
martpie: Also, we should allow for empty values: pseudo code: `if (value === "") formData.year = null`

src/views/ViewTrackDetails.tsx
```diff
@@ -115,6 +133,130 @@ export default function ViewTrackDetails() {
...
+              name="track"
+              type="number"
+              min="0"
+              step="1"
+              value={Number(formData.track.no)}
+              onChange={(e) => {
+                setFormData({
+                  ...formData,
+                  track: {
+                    no: Number(e.currentTarget.value),
```
martpie, line 181: Same for all these, they should allow empty values (that set the value to null), not 0 by default.

martpie (changes requested):
Overall looks great, thank you! Left comments.

For the CI failures, if you are using VSCode, I suggest you install the recommended extensions: they will give you linting and formatting warnings + format on save: https://github.com/martpie/museeks/blob/master/.vscode/extensions.json

martpie:
Thank you for the update, could you squash your commits and make sure master is the base branch? merge commits make it really hard to review as a the commits [from master are duplicated](https://github.com/martpie/museeks/pull/791/files).

igorer88:
Ok, I did a rebase and I think I broke it further. hahahha

imagen

I did a bun install But it' s still missing that macro error.

imagen

igorer88:
I'm starting to think that I should do a new fork hahahaha

martpie:
> I did a bun install But it' s still missing that macro error.

Try `cargo Install --path src-tauri` from the root of the repo? And make sure you got the latest version of rustup setup?

igorer88:
> Thank you for the update, could you squash your commits and make sure master is the base branch? merge commits make it really hard to review as a the commits [from master are duplicated](https://github.com/martpie/museeks/pull/791/files).

I've never done commit squash before, so let me investigate how to do it.

martpie:
https://www.git-tower.com/learn/git/faq/git-squash is a good resource :)

martpie:
If you use Github Desktop, it's fairly simple too: 

Screenshot 2024-10-06 at 12 46 36

Then I guess, fetch all, rebase from mater, and hopefully this shoud work. Force pushing to this branch is ok, or you can play with another branch to be safe.

igorer88:
Hi! @martpie sorry, I've been busy lately, had to travel and stuff. 
I'll do the squash and then I'll do another merge for the latest updates.

src/lib/utils-library.ts
igorer88: Done

igorer88:
hi! @martpie, the master branch is broken again? or maybe I need to check the last merge...

imagen

src/views/ViewTrackDetails.tsx
igorer88: done

src/views/ViewTrackDetails.tsx
igorer88: Agreed, done

src/views/ViewTrackDetails.tsx
igorer88: fixed to individual fields `track_no`, `track_of`, `disk_no`, `disk_of`...

src/generated/typings/index.ts
igorer88: changed back as it was before.

martpie:
Master should be passing(CI is), make sure you rebase and follow the setup instructions (bun install, bun tauri dev etc). Eventually, cargo test too if you are changing some types

martpie:
btw, are you using Limux Mac or Windows?

igorer88:
Mac

martpie:
Let me try to wrap this up, I'm finalising 0.20.0, and I'm doing a lot of chore/code revamps introducing a lot of conflicts

PR #795, opened 2024-10-31
SQLite

### LINKED ISSUE #793: Migrate DB to SQLite
I would like to migrate our DB (again!) to SQLite. Possible before the final release for Museeks Tauri.

SQLite would allow us to fully separate the back-end from the storage layer, as the back-end could be re-written in something else. Also, it would mean people could query the database themselves, etc.

I had a look at many crates to handle SQLite, from diesel to sea_orm to raw sqlite, and I was satisfied with none:

- diesel: codegen is ok, but platform-specific codegen is not.
- sea_orm: omg the doc is insanely unclear, especially with SQLite
- sqlx: learning another query language... meh, but why not
- tauri-plugin-sql: probably the simplest setup, uses sqlx under the hood

If anyone has experience with SQLite and rust, please wave :)

This is all assuming performances are ok.

martpie:
Fixes #793 

via the awesome `ormlite`.

- [x] MVP 
- [x] Fix tables not created on startup https://github.com/kurtbuilds/ormlite/issues/56
- [x] Add Indices
- [x] Add transactions/batching for:
  - [x] database reset (or use raw SQL)
  - [x] inserting tracks
- [x] Implement `reset`
- [x]  Check that everything works


martpie:
Got a working version, 4000 tracks load in 50ms. Insertion and deletions don't need transactions (inserting all 4000 tracks sequentially is almost instant).

Looking good so far :) Some weird stuff though with ormlite stuff not well exposed, but so far so good.

martpie:
Erratum: inserting 4000 tracks take 2 seconds, which is good enough. Deleting and vacuuming takes 5 sevconds clearing the records one by one, so I just used one statement to delete everything, which is instant, but less type safe

PR #802, opened 2024-12-03
Artists View

### LINKED ISSUE #130: More views than just TracksList: artists and albums
- [ ] AlbumsList (may require #86)
- [ ] ArtistsList 

_edit: July 12th 2023:_ the main challenge here is the DB design:
- either we completely re-think the DB, with relations, and everything (hard, error-prone, but easier to read and group data)
- we keep a single table for tracks, and make a lot of heuristic to know "which track belongs to which album/artist" (and try to avoid collisions)

martpie:
fixes #130

- [x] Artists list
- [x] Artist details
- [ ] Fix selection regression when clicking on a single item out of multiple selected ones
- [ ] Virtualized Lists ? Scroll helpers?
- [ ] Multi-artists
- [x] Jump to playing track
- [x] search

Screenshot 2024-12-03 at 15 58 15


igorer88:
It looks promising hehe reach out if I could help with something! 😁

martpie:
The main issue right now is the accessibility between multiple independent lists: arrow up / down etc should go from one album tracklist to the next etc. A bit challenging but I am getting there.

igorer88:
Talking about behavior... Have you checked [novajs.co](https://novajs.co/)? As they say it's _A collection of dependency-free React hooks_.
There are interesting hooks for keypress, localstorage, click detection, etc. they may come handy in this or any other feature.

martpie:
Been working on that for the past few days. I finally managed to fully decouple the view part from the behavior part on tracks list (selection, keyboard navigation etc), so now it's mostly about figuring out the UI + some of the special behavior, like what the search should do.

Getting there. Woof. 

https://github.com/user-attachments/assets/974d0ebd-76d0-4bb0-a747-428e21525b89

PR #805, opened 2024-12-04
Edit more ID3 tags (#732)

martpie:
Replaces #791 

Kinda impersonated @igorer88 for the commit, but I hope you don't mind.


martpie:
Don't worry, the rebase was pure hell anyway

martpie:
Thank you @igorer88, I'll leave you check if you want to tackle the ID3 edition on disk, but we should be good for now and it's already a win.

igorer88:
I like that option. I could really make use of that feature hehe. So I would continue. I have some of it ready in the git stash I think.

PR #816, opened 2025-01-04
Allow users to enable/disable blob playback

### LINKED ISSUE #815: Museeks 0.20.1 on MacOS 15.2: The operation is not supported. The track file could not be found. It may be due to a file move or an unmounted partition
I'm running Museeks Version 0.20.1 (20250102.162553) on MacOS 15.2, installed via [Homebrew](https://formulae.brew.sh/cask/museeks#default).

I'm unable to play my music library on Museeks, I'm getting this error message:

> The operation is not supported. The track file could not be found. It may be due to a file move or an unmounted partition.

This issue only happened yesterday when I updated my Homebrew applications, which included Museeks, with the following command:

`brew update && brew upgrade --no-quarantine`

On the previous version of Museeks, this wasn't an issue, so this is a bug on the new version of Museeks, and this isn't an issue with MacOS either, because I tested other music players, like VLC, and my music collection runs okay there, but not on Museeks.

Included in this bug report is a screenshot of the bug:

Screenshot 2025-01-04 at 4 50 49 PM

martpie:
There seems to be an issue with `convertFileSrc` on some mac, I have no idea why. This exposes the Linux playback (via blobs) as an option.

This is not a great UX solution, and does not allow me to understand what exactly is going wrong, but it should at least unblock users (hopefully). 

Maybe fix #815


martpie:
Better fix in #923

PR #833, opened 2025-03-11
Tanstack Router

martpie:
Attempt at using the more advanced react-router v7 features but omg y so complex though.

I really don't like giving ownership or vite to react-router, and may migrate to a completely different solution.


martpie:
I like it, it's a good change. I've been using `react-router` since... v2? But I'm sorry, those are too many breaking changes, and the framework +SPA setup is not seamless enough.

PR #849, opened 2025-06-05
Startup Improvements + Misc

martpie:
- [x] Fix Linux decorations being unresponsive 
- [ ] Prerender the root skeleton
- [x] Binary failing due to migrations not found
- [ ] Fix Icon on Linux

### DIFF
--- a/index.html
+++ b/index.html
@@ -13,4 +13,17 @@
       <!-- Here goes black magic -->
     </div>
   </body>
+
+  <style>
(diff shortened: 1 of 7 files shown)


martpie:
The fix for decorations bring too many drawbacks. If anyone is reading this, force the x11 rendering for Museeks for now with `GDK_BACKEND=x11 museeks`

PR #863, opened 2025-06-17
Oxlint

martpie:
Just playing. 

My two cents:

- Some rules are cool and enforce stylistic rules
- Generally, it feels more annoying that useful, and Biome's dedication to correctness instead of style may be the better design


martpie:
OXLint seems very petty in some of their recommended rules, Biome opinions are much more sane and do not hinder speed of development, so we'll stick with Biome.

PR #865, opened 2025-06-23
FlatHub and "Modern" Linux

martpie:
- [x] Need to go back to `npm` for better source generation for Flatpak
- [x] Need to drop support for Ubuntu < 24.04 (maybe, or just for flatpak, but that complexifies the CI)
- [ ] Ensure everything work
- [ ] Test builder manifest to build from source


martpie:
Will do the flathub checks later.

PR #871, opened 2025-06-27
added Russian translation

niya3:
Added Russian translation.

It's my first translation via `PO` files (and so with `lingui`) so i'm not sure about plural forms

Also I haven't build the application with the new translation, is it necessary?

### DIFF
--- a/src/translations/languages.ts
+++ b/src/translations/languages.ts
@@ -17,6 +17,12 @@ export const NON_DEFAULT_LANGUAGE = [
     code: 'fr',
     contributors: ['Pierre de la Martinière'],
   },
+  {
+    label: 'Русский',
+    englishLabel: 'Russian',
+    code: 'ru',
+    contributors: ['ildar nizamov'],
+  },
 ];
 
 export const ALL_LANGUAGES = [DEFAULT_LANGUAGE, ...NON_DEFAULT_LANGUAGE];
--- a/src/translations/ru.po
+++ b/src/translations/ru.po
@@ -1,0 +1,608 @@
+msgid ""
+msgstr ""
+"POT-Creation-Date: 2025-06-27 18:25+0300\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=utf-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"X-Generator: @lingui/cli\n"
+"Language: ru\n"
+"Last-Translator: Ildar Nizamov\n"
+"Plural-Forms: nplurals=3; plural=(n%10==1 && n%100!=11) ? 0 : ((n%10>=2 && n"
+"%10<=4 && (n%100<10 || n%100>=20)) ? 1 : 2);\n"
+
+
+#: src/components/AudioOutputSelect.tsx:53
+msgid "Loading devices..."
+msgstr "Загружается список устройств..."
+
+#: src/components/AudioOutputSelect.tsx:61
+msgid "Could not get audio output devices"
+msgstr "Не удалось получить список устройств вывода звука"
+
+#: src/components/AudioOutputSelect.tsx:69
+msgid "No audio devices found"
+msgstr "Звуковые устройства не обнаружены"
+
+#: src/components/ButtonRepeat.tsx:18
+msgid "Repeat"
+msgstr "Повтор"
+
+#: src/components/ButtonShuffle.tsx:12
+msgid "Shuffle"
+msgstr "Шафл"
+
+#: src/components/Cover.tsx:32
+msgid "Album cover"
+msgstr "Обложка альбома"
+
+#: src/components/DropzoneImport.tsx:52
+msgid "{skippedItemsCount, plural, one {# invalid item ignored} other {# invalid items ignored}}"
+msgstr "{skippedItemsCount, plural, one {# некорректный объект пропущен} few {# некорректных объекта пропущено} many {# некорректных объектов пропущено}}"
+
+#. placeholder {0}: folders.length
+#: src/components/DropzoneImport.tsx:63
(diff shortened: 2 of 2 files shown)


martpie:
Thank you for contributing! Regarding the plural form, I will need to have a look, but you can also try it by yourself running `bun tauri dev`, this will launch the app with the added translation which you can change in the settings :) 

You can ignore the macOS build failure btw.

martpie:
It seems that for Russian, you need to provide: one, few, many, other, as described here: https://cldr.unicode.org/index/cldr-spec/plural-rules.

So something like: `{numBooks, plural, one {# <noun>} few {# <noun>} many {# <noun>} other {# <noun>}} `

See also https://lingui.dev/guides/plurals

niya3:
Thank you! Yes, I've learned this rules the hard way :) russian translation seems finished to me, but I've been forced to add few untranslated lines in French - see dab72cb8..15c4029ab please

src/translations/ru.po
```diff
@@ -39,12 +39,12 @@ msgstr "Обложка альбома"
 
 #: src/components/DropzoneImport.tsx:52
 msgid "{skippedItemsCount, plural, one {# invalid item ignored} other {# invalid items ignored}}"
-msgstr "{skippedItemsCount, plural, one {# некорректный объект пропущен} few {# некорректных объекта пропущено} many {# некорректных объектов пропущено}}"
+msgstr "Пропущен{skippedItemsCount, plural, one { # некорректный объект} few {ы # некорректных объекта} many {о # некорректных объектов}}"
 
 #. placeholder {0}: folders.length
 #: src/components/DropzoneImport.tsx:63
 msgid "{0, plural, one {# folder added to the library} other {# folders added to the library}}"
```
niya3, line 46: btw, it seems to me, that `msgid_plural` and `msgstr[0..2]` template strings via mulltiple lines are much more readable than one-liners with `one`, `few`, `many`, `other` magic-words. One-liner are really pretty for one-word-type-of-strings, but they're hard to maintain in long enough lines like this one

niya3: I'm not sure how to use placeholder properly, but it may look kinda like here: https://github.com/nagyist/wp-project-manager/blob/8defe26561a566fc6cfefa32554f813830d1acb0/languages/cpm-cs_CZ.po#L665-L669

e.g.

```suggestion
msgid "{0} folder added to the library"
msgid_plural "{0} folders added to the library"
msgstr[0] "В библиотеку добавлена {0} папка"
msgstr[1] "В библиотеку добавлены {0} папки"
msgstr[2] "В библиотеку добавлено {0} папок"
```

and it may require some rework in the code, i think

niya3 (commented):
suggesting `msgid_plural`

src/routes/settings.ui.tsx
```diff
@@ -62,7 +62,7 @@ function ViewSettingsUI() {
       <Setting.Section>
         <Setting.Select
           label={t`Tracks density`}
-          description={t`Change the default view when starting the application`}
+          description={t`Change the tracks spacing`}
```
martpie, line 65: lel x_x

src/translations/fr.po
```diff
@@ -131,6 +131,10 @@ msgstr "voir plus"
 msgid "search..."
 msgstr "rechercher..."
 
+#: src/components/SideNavLink.tsx:59
+msgid "Rename"
+msgstr ""
```
martpie: You can put `Renommer` here :)

src/translations/fr.po
```diff
@@ -483,9 +489,8 @@ msgid "Tracks density"
 msgstr "Densité visuelle des pistes"
 
 #: src/routes/settings.ui.tsx:65
-#: src/routes/settings.ui.tsx:85
-msgid "Change the default view when starting the application"
-msgstr "Changer la vue par défaut au démarrage de l'application"
+msgid "Change the tracks spacing"
+msgstr ""
```
martpie: `Adapter l'espace entre les pistes`

src/translations/ru.po
martpie: Looks indeed simpler, but it's not really something I control, and it seems `lingui` is doing it that way. I will add something in the documentation about it and the plural form for each language.

martpie (approved):
Thank you for catching the few leftovers/mistakes :D

I left a few nits, you can correct them or I can do it myself, as you wish.

src/translations/languages.ts
```diff
@@ -17,6 +17,12 @@ export const NON_DEFAULT_LANGUAGE = [
     code: 'fr',
     contributors: ['Pierre de la Martinière'],
   },
+  {
+    label: 'Русский',
+    englishLabel: 'Russian',
+    code: 'ru',
+    contributors: ['ildar nizamov'],
```
martpie, line 24: Eventually, capitalize your name, but it's up to you :)

martpie (commented):
image

I see a menu item is on two line, if you want, feel free to add a few pixels to the menu width so that it stay on one line

src/translations/ru.po
niya3: Nice :)

I've just found in `lingui` documentation hint to getting `msgid_plurar`, https://lingui.dev/ref/catalog-formats#po-gettext

src/translations/languages.ts
niya3: I'd prefer this style, if it's ok to you :)

niya3:
> I see a menu item is on two line, if you want, feel free to add a few pixels to the menu width so that it stay on one line

Yes, I've tried it but it didn't work – unfortunately, CSS is a foreign language to me :) feel free sending the patch, I would prefer one-line view too

martpie:
I think we should be good to go, I will merge this and punish a release when I can, I have another fix to do before the next patch, thank you very much!

PR #882, opened 2025-07-09
E2E Tests via Vitest Browser

martpie:


martpie:
Quite happy with this setup, this will serve as the base for other E2E tests. Some things like the setup probably need to be abstracted for each though.

PR #894, opened 2025-08-04
Playwright

martpie:
Another attempt at E2E tests.


martpie:
Not possible to mock modules, so it's impossible to test the app reliably. Closing.

PR #906, opened 2025-09-18
feat: support to translate in traditional chinese

Mukvin:
As description, add the traditional chinese for translation


martpie:
Awesome, thank you very much! I want to add another language soon, then I'll release that in a patch :)

Issue #909, opened 2025-09-24
Large files playback stutters on Linux


raphial:
Hello! Glad to see there has been a lot of positive updates since I last used Museeks! However, I've run into a new issue on my Linux Mint 22.1 (kernel build6.8.0-84-generic, Museeks 0.23.1)  where audio seems to "skip" as if it's distorted or has playback issues, I'm not sure. It tends to skip every other second or so. I don't have this issue on other audio programs, only Museeks.

I have my music mounted in its own folder from over the network via Samba on my NAS server; don't know if this helps but it hasn't been an issue before.

martpie:
Hello there! 

I have tested Museeks on Mint 22.1 a couple of weeks ago and things seemed to work fine. Can you make sure your system is up to date?

How does your Samba setup look like exactly? Samba is not directly supported to load files, but I suspect this could be the issue, with Museeks not streaming enough data from the file due to the Samba setup.

I have a few ideas I can try and eventually give you a test binary and you can tell me if it's better.

yannick-1:
I'm seeing the same issue with my music on the NAS. Sometimes the program can hang for quite a bit. This used to work fine.

martpie:
Can you share your exact setup so I can try to reproduce it? Is it a network drive or something else? How did you add the music to the library etc?

yannick-1:
The NAS is a raspberry pi 4 running open media vault. The drive is connected via USB and uses NFS (3, 4, 4.1, 4.2).
The drive is mounted with ```sudo mount 192.168.0.220:/export/NAS /mnt/;```. 
Music is added through settings with folder ```/mnt/Music/```

superSTA101:
I have been experiencing the same problem but it occurs no matter where the files are located on the system
I'm on Arch Linux 6.16.10-zen1-1-zen my museeks version is 0.23.0

martpie:
To anyone facing this, can you try previous versions of Museeks and tell me if it happens there too?

Also, you can also try Museeks 0.13.1, which was based on Electron and should probably behave differently.

Assuming the issue is because of the drop of electron for a more lightweight solution, I have an idea what could cause this, but I need to spend some time experimenting there. TLDR is the way we load audio files on Linux is different compared to Windows/Mac due to an upstream issue, so ill have to see if I can find another workaround.

trainedmoose:
Museeks 0.13.1 seems to work fine without any audio skips.

Zorin OS 17.3 (using x86_64 AppImage)

[...]

Issue #912, opened 2025-10-14
App crashing on Wayland


BlueManCZ:
Hello, I'm not able to start the app on Wayland session. What is current status of Wayland support? Should it work or am I doing something wrong? Using the latest release 0.23.1 in Hyprland, glib-2.84.3.

```
museeks                                                                                               1 ↵
[init] Initialization successful
[file_associations] Handling opened files at startup: ["museeks"]
[file_associations] Found those valid paths: []
[2025-10-14][10:44:42][INFO][tracing::span] perform;
[2025-10-14][10:44:42][INFO][tracing::span] authenticate;
[2025-10-14][10:44:42][INFO][zbus::connection::handshake::common] write_command; command=Auth(Some(External), Some([49, 48, 48, 48]))
[2025-10-14][10:44:42][INFO][zbus::connection::handshake::common] write_commands; commands=[Auth(Some(External), Some([49, 48, 48, 48]))] extra_bytes=None
[2025-10-14][10:44:42][INFO][tracing::span] read_command;
[2025-10-14][10:44:42][INFO][zbus::connection::handshake::common] read_commands; n_commands=1
[2025-10-14][10:44:42][INFO][tracing::span] send_secondary_commands;
[2025-10-14][10:44:42][INFO][zbus::connection::handshake::common] write_commands; commands=[NegotiateUnixFD, Begin] extra_bytes=Some([108, 1, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 109, 0, 0, 0, 1, 1, 111, 0, 21, 0, 0, 0, 47, 111, 114, 103, 47, 102, 114, 101, 101, 100, 101, 115, 107, 116, 111, 112, 47, 68, 66, 117, 115, 0, 0, 0, 2, 1, 115, 0, 20, 0, 0, 0, 111, 114, 103, 46, 102, 114, 101, 101, 100, 101, 115, 107, 116, 111, 112, 46, 68, 66, 117, 115, 0, 0, 0, 0, 3, 1, 115, 0, 5, 0, 0, 0, 72, 101, 108, 108, 111, 0, 0, 0, 6, 1, 115, 0, 20, 0, 0, 0, 111, 114, 103, 46, 102, 114, 101, 101, 100, 101, 115, 107, 116, 111, 112, 46, 68, 66, 117, 115, 0, 0, 0, 0])
[2025-10-14][10:44:42][INFO][zbus::connection::handshake::client] receive_secondary_responses; expected_n_responses=1
[...]

PR #923, opened 2026-02-10
Add local HTTP server for audio streaming on Linux

BlueManCZ:
This commit introduces a local HTTP server for streaming audio on Linux as a workaround for WebKitGTK's asset protocol limitations with media streaming and Range requests. Includes a `stream_server` module and necessary changes to TS and Tauri integration.

I asked the Claude Opus AI agent to analyze this issue, and it came up with a working solution. This is its own review:

> On Linux, Museeks uses WebKitGTK as its webview engine, which uses GStreamer for audio decoding. Tauri's asset:// protocol doesn't support proper media streaming on WebKitGTK (https://github.com/tauri-apps/tauri/issues/3725). The previous workaround fetched entire audio files as blobs and created blob URLs, but GStreamer can't properly seek within blob URLs, causing:
> 
>   - FLAC: plays initial buffered data then stops (no seeking for more data)
>   - MP3: VBR seeking is broken, causing random position jumps
> 
>   Fix
> 
>   Replaced the blob URL workaround with a local HTTP server (using axum) that serves audio files with proper Content-Type, Accept-Ranges, and HTTP Range request support. GStreamer's HTTP source element (souphttpsrc) handles regular HTTP streaming perfectly.

I didn't test these changes on Mac or Windows, but on Linux this is night and day type of improvement.

Fixes: #909, #912 any maybe some other related playback issues?


martpie:
omg that's amazing, I was looking into this right now.

IIRC, I think this issue was also happening on mac with some files but I cannot reproduce it, I need to double check. If that's the case, we may want to open this via a custom setting and for all operating systems.

Let me try this branch once I get home.

martpie:
Just tested it, it works amazingly, thanks so much for doing that. I was really afraid I´d need to move the playback to the Rust side (tbh, maybe I'll still have to do it one day).

I'll review the code once I have a bit more time. You can ignore the macOS failure, it's expected as it uses my own certificate to bundle the app.

BlueManCZ:
>I was really afraid I´d need to move the playback to the Rust side

If you want me to prompt this and share what Opus would came up with, we can do that. :smile:

martpie:
hahahaha I have to say Claude has been working on a few PRs like #916

BlueManCZ:
I see, nice! :+1:

src-tauri/src/plugins/stream_server.rs
```diff
@@ -0,0 +1,184 @@
...
+use tauri::Runtime;
+use tokio::io::{AsyncReadExt, AsyncSeekExt};
+
+#[derive(Deserialize)]
+struct StreamParams {
+    path: String,
+}
+
+fn content_type_for_extension(ext: &str) -> Option<&'static str> {
+    match ext {
```
martpie: Can we keep that in sync with https://github.com/martpie/museeks/blob/6656f952b54654dcac6a40ae7ed438988f7116b3/src-tauri/src/libs/database.rs#L12 ?

Basically, if an extension is added there, this function should break saying a case is not covered. This may require changing the type of the constant though.

src-tauri/src/plugins/stream_server.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    Some((start, end))
+}
+
+fn build_response(status: StatusCode, content_type: &str, file_size: u64, body: Vec<u8>, range: Option<(u64, u64)>) -> Response {
+    let mut headers = AxumHeaderMap::new();
+    headers.insert(header::CONTENT_TYPE, content_type.parse().unwrap());
+    headers.insert(header::CONTENT_LENGTH, body.len().to_string().parse().unwrap());
+    headers.insert(header::ACCEPT_RANGES, "bytes".parse().unwrap());
+    headers.insert(header::ACCESS_CONTROL_ALLOW_ORIGIN, "*".parse().unwrap());
```
martpie: Not sure this header is required is it?

src-tauri/src/main.rs
```diff
@@ -29,7 +29,12 @@ fn main() {
         }
     };
 
-    tauri::Builder::default()
+    // On Linux, start a local HTTP server for audio streaming.
+    // WebKitGTK's asset protocol doesn't support media streaming properly.
+    #[cfg(target_os = "linux")]
+    let stream_server_port = plugins::stream_server::start();
```
martpie: any way this logic can be moved within the init of the plugin? I understand it's a sync init.

See how I did it for the DB:

https://github.com/martpie/museeks/blob/6656f952b54654dcac6a40ae7ed438988f7116b3/src-tauri/src/plugins/db.rs#L467

It may create weird rust errors though, if it's too hard to change or fix the errors correctly, feel free to leave them like that, it's okeeey-ish.

src-tauri/Cargo.toml
```diff
@@ -56,6 +56,9 @@ ts-rs = "12.0.1"
 uuid = { version = "1.20.0", features = ["v3", "v4", "fast-rng"] }
 walkdir = "2.5.0"
 
+[target.'cfg(target_os = "linux")'.dependencies]
+axum = "0.7"
```
martpie, line 60: genuine question, how heavy is this? are there other minimalist-yet-ergonomic http servers here? not a big deal.

martpie (commented):
A bunch of nits, I'm ready to merge it as-is if you don't have the bandwidth to handle those changes :)

src-tauri/src/plugins/stream_server.rs
BlueManCZ: I put this into the `database.rs` and derived `SUPPORTED_AUDIO_FORMATS`.

src-tauri/src/plugins/stream_server.rs
BlueManCZ: Yes, can be removed.

src-tauri/src/main.rs
BlueManCZ: Moved to plugin

src-tauri/Cargo.toml
BlueManCZ: Based on what Opus analyzed it is very light in this case. It only added 5 new crates to the lock file (`axum`, `axum-core`, `matchit`, `serde_path_to_error`, `httpdate`). Everything else axum depends on (`hyper`, `http`, `tokio`, `tower`, `serde`, `bytes`, `futures-util`, etc.) was already in the tree via Tauri. So the actual added weight is minimal.

The main alternatives would be:
- `warp` - similar weight (also built on hyper/tokio), but its filter-based API is less ergonomic for this use case
- `tiny_http` - truly minimal, but synchronous (doesn't use tokio), so it'd need its own thread pool for concurrent requests
- Raw `hyper` - zero extra crates, but significantly more boilerplate for routing, query parsing, and response building
- Raw `tokio::net::TcpListener` - zero deps, but manual HTTP parsing is error-prone

Given that axum reuses almost the entire existing dependency tree and only adds 5 small crates, it's a good trade-off for correct, maintainable code.

martpie:
Let me merge that, I think I may transform this into a Setting for people to opt-in, regardless of what OS they use, and there are a couple of nit I may do.

Thank you, hopefully I can release that this week, if not, when you can build from sources in the meantime haha (`npm run tauri build`).

PR #925, opened 2026-02-16
Switch from biome to oxc

martpie:
I don't know if I'll move forward with this, I'm just playing with it. There are things I like, things I don't.

Overall, Biome opinions feel nicer, everything make sense. On the other hand, `oxc` may have momentum (as it come from the creators of `vite`), so better integration with vitest, etc.


martpie:
screw it let's go

PR #935, opened 2026-03-06
Group all Compilations Albums within their own view

### LINKED ISSUE #933: Group compilation tracks into a virtual "Various" or "Compilations" artist.
_Hi, just a quick note to say that I really enjoy this app and am especially mindful of its simplicity. The feature request I describe here would not add clutter to the app, but is in my view one of few missing essential features for Museeks to be used as one’s main music player. Thanks for taking the time to read, and if this issue should be moved to discussions I can do this!_

### **Feature description:**
Support for « compilation » albums
- In artists view, instead of being ordered with the artist, tracks which are part of a compilation (tag COMPILATION == 1) are shown as part of a new « Compilations » entry at the bottom of the artist view, where they are ordered by album (chronologically or alphabetically)
- When viewing the compilations, next to each compilation album, an additional third column with the track’s artist is shown (between the track title and track length)
This is how compilations are managed in most music players, e.g. iTunes.

Image

### **Why this could matter to the user:**
Currently compilation albums don’t show as a single album.
This means that:
1. Compilation albums cannot be viewed or listened to as a single album, as they should be
2. The artist view is cluttered by numerous artists (sometimes less known), often with only one track

### **Why this wouldn’t clutter the app:** 
It is a background feature, invisible if one doesn’t have compilation albums in his/her music library, and can significantly clean up one’s library if so. Especially, it would organise albums they way they should be.

martpie:
Fixes #933 

- [x] Fix horizontal overlay on compilations view

image

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -94,17 +94,17 @@ src-tauri/               # Rust/Tauri backend
 
 **UI:**
 
-- Never call `invoke` directly from React components — use or extend the `bridge-*.ts` helpers in `src/lib`.
+- Never call `invoke` directly from React components — extend the `bridge-*.ts` helpers in `src/lib` (or re-use the same pattern).
 - `src/elements` are business-agnostic UI primitives. `src/components` and `src/routes` hold business logic.
 - Prefer URL state over Zustand global state unless persistence is required.
 - TypeScript types generated from Rust live in `src/generated` and should be committed — regenerate via `gen:types`.
-- After editing the configuration of a route, regenerating the route-tree must be done via the `build` script.
 
 ## Agent Operations Rules
 
-- After UI edits: run `bun run test:lint` (covers linting + type-check) and `bun run test:format` (formatting).
-- After Rust edits: run `cargo test` in `src-tauri`. Clippy is a bonus.
+- After UI edits: ensure `bun run test:lint` (covers linting + type-check) and `bun run test:format` (formatting) pass.
+- After Rust edits: ensure `cargo test` in `src-tauri` passes. Clippy is a bonus.
 - After modifying a Rust struct exposed via `ts-rs`: run `bun run gen:types` and commit the output.
+- After editing the configuration of a route, regenerating the route-tree must be done via the `build` script.
 - Never manually edit `src/generated` — always regenerate.
 - Don't fix pre-existing issues unrelated to the current task.
 - When uncertain, read nearby file patterns first, then ask.


martpie:
cc @Raphael0 I'll send you a .dmg when it's ready for feedback, you're on macOS right?

Raphael0:
Yes, I’m on the latest Tahoe version, happy to test it!

martpie:
Here's a link: https://drive.google.com/file/d/1gZRiACfRWL9cprrtIcwhAKoKYEC_Y_1O/view?usp=drivesdk

- you will need to full-rescan your library from the settings to enable it
- there is one layout issue that may happen with very long track+artist name, I'll fix it later

Raphael0:
Thanks, I’ll take a look later today

martpie:
I think we're good to go

Raphael0:
Well first that was a very good exercise to find out that some of my music wasn't tagged correctly :)

Three things:
- Clicking on the artist or album in the now-playing panel at the top leads to a "There is no such music in your library" error when the track is part of a compilation. When clicking on the album, it should open the album under the compilations pane. When clicking on the artist, the solution is more tricky! What happens now is that you get an error if it is (1) an artist that only appears in compilations. If it is (2) an artist that appears in a compilation but also appears in a "regular" album, the artist pane is opened (which I guess is good). I suppose that the best would be to either open the track's album in the compilation pane or deactivate the artist link for tracks of (2) artists, since they don't have a dedicated artist pane. The logic might be more finicky to implement.
- I would add the compilations pane at the bottom of the artist list rather than at the top, since they appear very prominently now, but that's just a preference
- I do think adding a "This track is part of a compilations album with multiple artists" checkbox in the track edit window to toggle the compilations tag would be a nice addition, especially for users that have not tagged their whole collection consistently and will help them debug if their collection displays strangely + edit it

With this, thanks for the super fast and great update!

Raphael0:
Also just to expand on the third item, I think that this would be useful for "non-power users" in particular (power users will know what is going on and might use a dedicated tagging software). A regular user might be confused by the way that the app classifies albums as compilations or not, and without this option they wouldn't have a way to find out why and how to correct it (most people will have music from many different sources, so inconsistent tagging can happen easily). It would add a certain "transparency" to the feature even though it does add a new setting which goes against simplicity. Just my two cents of course!

martpie:
> Clicking on the artist or album in the now-playing panel at the top leads to a "There is no such music in your library" error when the track is part of a compilation.

Right, good catch

> I would add the compilations pane at the bottom of the artist list rather than at the top, since they appear very prominently now, but that's just a preference

I'm not sure about this yet. Ideally, at the top would be other options like "all artists" etc. But true, in the meantime we could put it at the bottom. No strong opinion here.

> I do think adding a "This track is part of a compilations album with multiple artists" checkbox in the track edit window to toggle the compilations tag would be a nice addition, especially for users that have not tagged their whole collection consistently and will help them debug if their collection displays strangely + edit it

Right.

Should be fairly simple to implement those changes.

PR #963, opened 2026-04-18
Add an album list/view

adrhumphreys:
This just adds a basic album view, I found that often there are tracks in albums by artists other than the main (e.g. artist x + artist y) so the artist view would split things up and have the odd track missing from an album.

image


martpie:
Hey, really sorry for the late reply, I'll review this when I can, sorry I've been a bit busy these days.

adrhumphreys:
no worries, and thanks for the update

